PR #244, opened 2026-06-15
fix file read

bomzheg:


shvatka/core/services/game.py
```diff
@@ -163,22 +163,22 @@ async def get_full_game(
     dao: GameByIdGetter,
     org_dao: OrgByPlayerGetter,
```
bomzheg: `org_dao` is now an unused parameter — the body switched to `check_can_view_scenario(game, identity)` and no longer touches it. All four callers (`game_scn/handlers.py:109`, `editor_interactors.py:62`, `game_publish/handlers.py:55`, `organizer.py:28`) still pass `org_dao=dao.organizer`. Not a bug, but worth dropping the parameter (and the now-redundant `OrgByPlayerGetter` import here) and updating those call sites to keep the signature honest.

---

shvatka/tgbot/services/identity.py
```diff
@@ -99,6 +101,18 @@ async def get_full_team_player(self) -> dto.FullTeamPlayer | None:
         self.cache["full_team_player"] = team_player
         return team_player
 
+    async def get_org(self, game: dto.Game) -> dto.Organizer | None:
```
bomzheg, line 104: This `get_org` implementation is byte-for-byte identical to the one added in `shvatka/api/dependencies/auth.py`. Both depend only on `self.dao.organizer.get_by_player_or_none(...)` and `self.get_required_player()`, so it could be hoisted into the base `IdentityProvider` (or a small mixin) to avoid the two copies drifting apart later.

Minor: `await self.get_required_player()` is evaluated twice on the non-author path. `get_required_player` is cached so it's cheap, but binding it to a local once reads cleaner.

---

bomzheg (commented):
## Review — `fix file read` (PR #244)

Solid, well-scoped change. It centralizes "who may read scenario files" behind `IdentityProvider.get_org()` and makes `check_file_meta_can_read` consistent with `check_can_view_scenario`, so a secondary organizer holding `view_scenario` (and the game author, via `PrimaryOrganizer`) can now read files of an in-progress game instead of being rejected. That's the right fix and removes the old `get_file_metas`/`get_game_file_metas` duplication.

I verified locally:
- All callers of the removed `get_game_file_metas` and `get_file_content` are gone — no dangling imports (`grep` clean), and all changed modules parse.
- `dto.Game.author` is a required field, so `game.author.id` / `PrimaryOrganizer(player=game.author, ...)` inside `get_org` can't `AttributeError`.
- The `org_dao` constructor arg dropped from `AllGameKeysReaderInteractor` / `GameScenarioTransitionsInteractor` is safe — they're wired via `provide(...)` in `di/interactors.py`, which resolves from the `__init__` signature.

### What I like
- Per-game `get_org` caching keyed by `game.id`, including caching negative (`None`) results, with the game author short-circuited to `PrimaryOrganizer`.
- Dedup of guids (`seen` set) is now applied on the `get_game_package` path too, which previously used the non-deduplicating variant.

### Suggestions (non-blocking)
1. **Dead parameter** — `get_full_game` still takes `org_dao` but no longer uses it (inline comment). Worth removing along with its callers.
2. **Duplicated `get_org`** — identical in `auth.py` and `tgbot/services/identity.py`; consider hoisting to the base `IdentityProvider` (inline comment).
3. **Test coverage** — for a PR titled "fix file read", a regression test for "a `view_scenario` org can read a file of an incomplete game" (and the negative: a plain org cannot) would lock in the behavior. Nothing in `tests/` currently exercises `check_file_meta_can_read` / `get_file_metas`.

No correctness blockers found. 👍

---

PR #276, opened 2026-07-01
Fall back to sending hint by content when file_id is missing

bomzheg:
## Problem

`HintSender.send_hint` sends a hint by `file_id` first and falls back to sending by content on `TelegramAPIError`. Two gaps:

1. **No fallback when `file_id` is null.** When a file's `file_id` was never set (nullable column), `resolve_link` produced a view with `file_id=None`. Passing `photo=None` (etc.) to aiogram raises a *validation* error, not a `TelegramAPIError`, so the existing content fallback was never triggered and the hint failed to send.
2. **File_id was never renewed.** When a hint *was* sent by content, telegram returns a fresh `file_id`, but it was discarded — so every subsequent send repeated the expensive content upload.

## Changes

- `send_hint` now detects a missing `file_id` up front (`_is_file_id_missing`) and goes straight to sending by content, keeping the `TelegramAPIError` fallback for the case where a present `file_id` is rejected by telegram.
- After sending by content, `_renew_file_id` parses the returned message for the new `file_id` and persists it. As guessed, this uses a **dedicated session (separate transaction)** via the injected `async_sessionmaker`, so the renewal is committed independently of the request's transaction.
- Renewal is best-effort: the hint is already delivered, so any failure while renewing is logged and swallowed rather than propagated.
- `HintSender` now takes a `pool: async_sessionmaker[AsyncSession]` dependency (auto-wired by dishka; test fixture updated accordingly).

## Tests

- `test_send_by_content_when_file_id_missing` — a file with a null `file_id` is sent by content with no failed `file_id` attempt.
- `test_renew_file_id_after_send_by_content` — after the `file_id` send fails and content is sent, the DB `file_id` is updated to the value returned by telegram.

---
_Generated by [Claude Code](https://claude.ai/code/session_019zHj5eQ1hUfQt44m6s3jSV)_

### DIFF
--- a/shvatka/tgbot/views/hint_sender.py
+++ b/shvatka/tgbot/views/hint_sender.py
@@ -63,7 +63,7 @@ class HintSender:
             try:
                 return await method(chat_id=chat_id, **hint_link.kwargs())
             except TelegramAPIError:
-                logger.warning("cant send hint by file_id %s", hint_link)
+                logger.warning("cant send hint by file_id %s", hint_link, exc_info=True)
         return await self._send_by_content(method, hint_container, chat_id)
 
     async def _send_by_content(
--- a/shvatka/tgbot/views/hint_sender.py
+++ b/shvatka/tgbot/views/hint_sender.py
@@ -62,8 +62,8 @@ class HintSender:
         else:
             try:
                 return await method(chat_id=chat_id, **hint_link.kwargs())
-            except TelegramAPIError:
-                logger.warning("cant send hint by file_id %s", hint_link, exc_info=True)
+            except TelegramAPIError as e:
+                logger.warning("cant send hint by file_id %s", hint_link, exc_info=e)
         return await self._send_by_content(method, hint_container, chat_id)
 
     async def _send_by_content(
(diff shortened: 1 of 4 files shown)


shvatka/tgbot/views/hint_sender.py
```diff
@@ -45,12 +57,47 @@ def method(self, type_content: enums.HintType) -> Callable[..., Awaitable[Messag
...
-            hint_content = await self.resolver.resolve_content(hint_container)
-            return await method(chat_id=chat_id, **hint_content.kwargs())
+            tg_link = parse_message(message)
+            if tg_link is None:
+                return
+            await self.file_info_dao.update_file_id(guid, tg_link.file_id)
+            await self.file_info_dao.commit()
+        except Exception:
+            # renewing file_id is best-effort and must never break sending
+            logger.exception("cant renew file_id for hint with guid %s", guid)
```
bomzheg: add traceback logging

bomzheg: The line-100 handler already logs a traceback — it uses `logger.exception(...)`, which sets `exc_info=True` implicitly. The one place that was missing a traceback was the `except TelegramAPIError` fallback (the `logger.warning("cant send hint by file_id ...")`), so I added `exc_info=True` there in 2247be6. Let me know if you actually meant a different spot.

---

PR #277, opened 2026-07-01
Add email registration, login and account linking

bomzheg:
## What

Adds **email** as a login identity alongside Telegram. Players can now register and log in by email, and link email ⇄ telegram in **both directions**. The password stays a single shared credential on `Player`, and email verification is **required** before email login works (can be relaxed later).

## Design

Email is modeled as a third linked identity on `Player`, mirroring the existing `User` (tg) / `ForumUser` pattern — a one-to-one `EmailAccount`. All email flows are dataclass **interactors** registered in DI (`EmailInteractorProvider`), so routes and bot handlers just receive them via `FromDishka[...]`.

## Changes

**Core & storage**
- `EmailAccount` model + DTO, new `emails` table (alembic migration `e5f6a1b2c3d4`), unique on `email` and `player_id`
- `EmailAccountDao` (rdb) + redis-backed `EmailConfirmationStore` for 6-digit confirmation codes
- `core/services/email.py`: `EmailRegisterInteractor`, `EmailLinkInteractor`, `EmailConfirmInteractor`, `EmailResendInteractor`
- `EmailSender` interface with `aiosmtplib` SMTP sender and a console fallback (logs the code) selected by a new **optional** `MailConfig`
- shared `PasswordHasher` (bcrypt) — `AuthProperties` now delegates to it too, so all password hashing goes through one place

**Web API** (`routes/auth.py`)
- `POST /auth/register/email` — register a new player by **username** + email + password (username validated & unique)
- `POST /auth/login/email` — login by verified email + password
- `POST /auth/email/confirm` — confirm with the emailed code
- `POST /auth/email/resend` — request a new confirmation code (silent on unknown email)
- `POST /auth/email/link` — attach an email to the logged-in player (no password manipulation — just adds the email)
- `POST /auth/link/tg` — attach a telegram account to an email-authenticated player (via the tg login widget)

**Telegram bot**
- Profile dialog flow to attach and confirm an email

**Tests**
- Unit coverage for all email interactors (register / duplicates / invalid input / confirm / link / resend)

## Config

New optional `mail:` section (see `config_dist/config.yml`). When disabled, confirmation codes are only written to the log, so local/dev and tests work without SMTP credentials. For ~100 users a Gmail/Yandex app-password mailbox over SMTP is sufficient.

## Notes / follow-ups

- Email verification is currently mandatory before email login; making it optional later is a small change.
- Rate-limiting of the register/login/confirm endpoints is not included here.
- A tg-only player without a password must set one (existing `PUT /users/me/password` / profile flow) before email login works after linking.


shvatka/core/interfaces/dal/email.py
```diff
@@ -0,0 +1,43 @@
+import typing
+
+from shvatka.core.interfaces.dal.base import Committer
+from shvatka.core.models import dto
+
+
+class EmailAccountDao(typing.Protocol):
+    async def get_by_email(self, email: str) -> dto.EmailAccount | None:
+        ...
```
bomzheg: why sometimes you write pass, sometimes raise NotImplementedException and sometime ellipsis? 
make exceptions here pls

shvatka/core/services/email.py
```diff
@@ -0,0 +1,109 @@
...
+    if normalized is None:
+        raise exceptions.EmailInvalid(text=f"invalid email {email}")
+    return normalized
+
+
+def _generate_code() -> str:
+    return "".join(secrets.choice("0123456789") for _ in range(CODE_LEN))
+
+
+async def register_by_email(
```
bomzheg: why this and another function in this module is not interactors? it's better to register it and get all nedded from DI (if register as register_email = provide(MyInteractor) we dont need to have manual deps management

shvatka/core/services/email.py
```diff
@@ -0,0 +1,109 @@
...
+    hasher: PasswordHasher,
+    store: EmailConfirmationStore,
+    sender: EmailSender,
+) -> None:
+    """Attach an email to an already existing player (e.g. a telegram user)."""
+    email = _normalize_or_raise(email)
+    if await dao.is_email_occupied(email):
+        raise exceptions.EmailAlreadyExist(text=f"email {email} already occupied")
+    if password is not None:
+        await dao.set_password_if_absent(player, hasher.hash(password))
```
bomzheg: what usecase is this? user add new password but we dont change it if he already have one? meh.

i guess add email should just add mail. no password manipulations. or explain why should we add

shvatka/infrastructure/crypto/hasher.py
```diff
@@ -0,0 +1,12 @@
+from passlib.context import CryptContext
+
+
+class BcryptPasswordHasher:
```
bomzheg, line 4: if so another password haches should also hashed by this class

shvatka/infrastructure/db/dao/rdb/email.py
```diff
@@ -0,0 +1,94 @@
...
+        db_account = await self._get_by_email_or_none(email)
+        return db_account.to_dto() if db_account is not None else None
+
+    async def is_email_occupied(self, email: str) -> bool:
+        result = await self.session.scalars(
+            select(models.EmailAccount).where(func.lower(models.EmailAccount.email) == email)
+        )
+        return result.one_or_none() is not None
+
+    async def create_player_for_email(self, email: str, hashed_password: str) -> dto.Player:
```
bomzheg: let's make required username for registration via email

shvatka/core/interfaces/dal/email.py
bomzheg: Fixed in 1f78bcb7 — all protocol stubs now use `raise NotImplementedError`, matching the style in `core/interfaces/dal/player.py`. Also collapsed the two identical marker protocols (`EmailRegistrar`/`EmailConfirmer`) into a single `EmailDao`.

---

shvatka/core/services/email.py
bomzheg: Done in 1f78bcb7 — converted to dataclass interactors (`EmailRegisterInteractor`, `EmailLinkInteractor`, `EmailConfirmInteractor`, `EmailResendInteractor`), registered in a new `EmailInteractorProvider` (`infrastructure/di/mail.py`, added to `get_providers`). API routes and bot dialog handlers now just take the interactor via `FromDishka[...]` — no manual dependency wiring at call sites.

---

shvatka/core/services/email.py
bomzheg: Agreed, removed in 1f78bcb7 — linking an email now just adds the email, no password manipulation (`set_password_if_absent` is gone, along with the bot's password window and the `password` field on the link endpoint). The original intent was that a tg-only player has no password, so email login would be impossible until one is set — but that's better solved by the existing password-setting flow (`PUT /users/me/password`), not implicitly during linking.

---

shvatka/infrastructure/crypto/hasher.py
bomzheg: Done in 1f78bcb7 — `AuthProperties` no longer owns its own `CryptContext`; it now receives the shared `PasswordHasher` via DI and delegates `verify_password`/`get_password_hash` to it, so every password in the system goes through the same hasher (`api/password_hash.py` script updated too).

---

shvatka/infrastructure/db/dao/rdb/email.py
bomzheg: Done in 1f78bcb7 — `POST /auth/register/email` now requires `username` in the body; it's validated with the same rules as elsewhere (`validate_new_username`) and checked for uniqueness (409 on conflict). The auto-generated `id{N}` fallback is gone.

---

shvatka/core/games/interactors.py
```diff
@@ -82,7 +82,7 @@ def __init__(
     async def __call__(
         self, guid: str, game_id: int, identity: IdentityProvider
     ) -> VerifiableFileMeta:
-        user = await identity.get_required_user()
+        user = await identity.get_user()
```
bomzheg: inline it. we dont need to load it if will not use (and we dont use it in happy path) (yes, i now about caching and even so)

shvatka/core/models/dto/export_stat.py
```diff
@@ -25,25 +26,27 @@ class Player:
...
-            player_identity = PlayerIdentity.tg_user_id
-            return Player(tg_user_id=player_tg_id, identity=player_identity)
+            return Player(tg_user_id=player_tg_id, identity=PlayerIdentity.tg_user_id)
         elif player.has_forum_user():
             player_name = player.get_forum_name()
             assert player_name is not None
-            player_identity = PlayerIdentity.forum_name
-            return Player(forum_name=player_name, identity=player_identity)
+            return Player(forum_name=player_name, identity=PlayerIdentity.forum_name)
+        elif player.username is not None:
```
bomzheg: i guess it should be first (and mostly the one - username actual have any player, but ok, let's we have fallback for tg user and forum user identities)


for all reasons player.username should be first possible things for show. actual dto.Player already have name_mention where we have all needed logic. may be use it and remove in current module their own?

PR #301, opened 2026-07-22
Fix scenario re-upload, and files being stored empty after upload to telegram

### LINKED ISSUE #298: scenario can't be reuplod
```
shvatka-backend-1  |         +-+---------------- 1 ----------------
shvatka-backend-1  |           | adaptix.load_error.TypeLoadError: expected_type=None, input_value={'file_id': None, 'content_type': 'video'}
shvatka-backend-1  |           +---------------- 2 ----------------
shvatka-backend-1  |           | Exception Group Traceback (most recent call last):
shvatka-backend-1  |           |   File "/opt/venv/lib/python3.13/site-packages/adaptix/_internal/morphing/generic_provider.py", line 416, in optional_dt_loader
shvatka-backend-1  |           |     return loader(data)
shvatka-backend-1  |           |   File "<adaptix generated model_loader_TgLink>", line 77, in model_loader_TgLink
shvatka-backend-1  |           |     raise AggregateLoadError(
shvatka-backend-1  |           |     ...<2 lines>...
shvatka-backend-1  |           |     )
shvatka-backend-1  |           | adaptix.load_error.AggregateLoadError: while loading model <class 'shvatka.core.models.dto.hints.file_content.TgLink'> (1 sub-exception)
shvatka-backend-1  |           +-+---------------- 1 ----------------
shvatka-backend-1  |             | Traceback (most recent call last):
shvatka-backend-1  |             |   File "<adaptix generated model_loader_TgLink>", line 52, in model_loader_TgLink
shvatka-backend-1  |             |     f_file_id = loader_file_id(r_file_id)
shvatka-backend-1  |             |   File "/opt/venv/lib/python3.13/site-packages/adaptix/_internal/morphing/concrete_provider.py", line 481, in str_strict_coercion_loader
shvatka-backend-1  |             |     raise TypeLoadError(str, data)
shvatka-backend-1  |             | adaptix.load_error.TypeLoadError: expected_type=<class 'str'>, input_value=None
shvatka-backend-1  |             | Exception was caused at ['file_id']
shvatka-backend-1  |             +------------------------------------

```

bomzheg:
Three related fixes in the file-meta / upload area. The last one is an independent data-corruption bug found while testing this branch and can be split out if you'd rather merge it separately.

## 1. Scenario re-upload crashed on files without a telegram file_id (#298)

Re-uploading a scenario failed with:

```
adaptix.load_error.TypeLoadError: expected_type=<class 'str'>, input_value=None
Exception was caused at ['file_id']
```

A file uploaded via the web API has no telegram `file_id` yet — the column is nullable and `fill_all_file_id` backfills it lazily — but the scenario model required a `str`.

**Fix (per review):** `TgLink` keeps a required `file_id`; what is optional is the whole link. `file_id`/`content_type` are now plain fields on the file meta, and `tg_link` is a derived property returning `None` when there is no file_id:

```python
@property
def tg_link(self) -> TgLink | None:
    if self.file_id is None or self.content_type is None:
        return None
    return TgLink(file_id=self.file_id, content_type=self.content_type)
```

No `*FileMeta` carries a `TgLink | None` field, and no `assert file_id is not None` guards are needed. A file with no file_id yields `tg_link is None`, which the existing `if not file_meta.tg_link` upload path and the send-by-content fallback already handle.

This changes the scenario/zip format: `file_id` and `content_type` are written flat instead of nested under `tg_link`.

## 2. Scenarios exported before that change still load

Inlining made the old nested `tg_link` key unknown, so adaptix silently dropped it and old zips lost their `content_type` — then re-uploading them died on the `assert file_meta.content_type is not None` in `upload_to_tg`, i.e. exactly the flow #298 is about.

The loader now migrates the legacy shape to the inlined fields. The compatibility lives only in the retort recipe; the DTOs stay clean and dumping still emits the inlined form.

## 3. Files were stored empty after being uploaded to telegram

`BotFileGateway.put` read the content stream twice:

```python
if not file_meta.tg_link:
    await self.upload_to_tg(author, content, file_meta)   # content.read() drains it
saved_file = await self.storage.put(file_meta, content)   # reads b""
```

So files that had to be uploaded to telegram (those without a file_id) were written to disk as **0 bytes**, with the sha256 of the empty string and `mime_type=application/x-empty` — while the telegram upload itself succeeded, so the row got a valid `file_id` and looked healthy. Serving such a file returns `Content-Length: 0`.

This predates the branch, but #298 makes `file_id: null` legitimately common, so the path is hit more often now.

Fixed by buffering the content when it has to be uploaded, so telegram and the storage both get the whole file.

### Repairing already-broken files

They cannot heal on their own: the local file exists and just reads back empty, so the download-from-telegram fallback in `get()` never fires. Their content is still in telegram, though, and the empty sha256 identifies them exactly.

Added `shvatka-empty-file-repairer` (a sibling of `shvatka-file-id-updater`): it finds those files, downloads the content back from telegram and refreshes `sha256`/`mime_type`. Files without a file_id, and files telegram no longer has, are logged and left untouched so one dead file cannot abort the run.

## Tests

- `test_reupload_scenario_with_file_without_tg_file_id` — a scenario whose file has no file_id loads (#298).
- `test_deserialize_legacy_scenario_with_nested_tg_link` / `..._without_tg_file_id` — pre-inlining scenarios still load with both values preserved.
- `tests/unit/test_file_gateway.py` — content uploaded to telegram is still stored in full; verified this fails without the fix.

Fixes #298

https://claude.ai/code/session_01J8nEvUZ7njjDA4d7m7W6wZ


shvatka/infrastructure/clients/file_gateway.py
```diff
@@ -59,9 +59,12 @@ async def upload_to_tg(
         await msg.delete()
         tg_link = parse_message(msg)
         assert tg_link
+        assert tg_link.file_id is not None
```
bomzheg: oh. i created many types for do not assert is not None as much as possible. can we do it with as less optionals and union as possible?

i know this file staff became complex:(
may be you have any ideas

bomzheg: Good catch — and on your follow-up ("how do we know the hint type then?"): `content_type` lives inside `tg_link` in the scenario YAML, so dropping `tg_link` when there's no `file_id` would lose the type and force a format change. So I kept `file_id` optional only on the base `TgLink`.

To remove the asserts I declared `ParsedTgLink.file_id: str` (non-optional) — a link parsed from a live telegram message always has a `file_id`, and mypy honours the narrowing. That drops the `assert ... is not None` here in `upload_to_tg`, plus the ones in `hint_parser` and `_renew_file_id`. The only guard left is in `download_from_tg`, where a *stored* link genuinely may have nothing to download. Fixed in 2f70551.

---

PR #303, opened 2026-07-24
Pin puzzles and hints in team chats

### LINKED ISSUE #269: пинить подсказки в чат

bomzheg:
Closes #269

## What

Bot now pins what it sends to a team chat and unpins it when it goes stale:

| pinned | unpinned |
| --- | --- |
| level puzzle, time hints (`PinCategory.level`) | when the team goes to the next level (`send_puzzle`), and at the end of the game |
| bonus hints from a key or a timer (`PinCategory.bonus`) | when the game is finished for the team (`game_finished` / `game_finished_by_all`) |

All parts of a multipart hint are pinned, and so is the caption sent before them
("Уровень № 3", "Бонусная подсказка", ...).

Leftovers from a game that ended abnormally are unpinned in `prepare_game_view`,
when the chat is prepared for the next game.

## How

- `HintSender.send_hints` now returns every message it sent (caption first), instead of returning `None`.
- Pinning itself lives in a new component, `MessagePinner` (`shvatka/tgbot/views/pinner.py`).
  `BotView` feeds it the messages it got from `HintSender` and asks it to unpin a category
  when that category becomes outdated.
- The bot is not necessarily an admin in a team chat (or may lack the "pin messages" right),
  so every telegram call in the pinner is best-effort: an error is logged and never propagates
  into the game flow. The same holds for storage errors.
- Ids of pinned messages are stored per chat and category in redis by `PinnedMessageDao`
  (a list per key, with a 30-day TTL), so messages can still be unpinned after a bot restart.
  Wired via DI (`DAOProvider.pinned_message_dao`, `GameToolsProvider.message_pinner`) - no new
  `HolderDao` property.

### Bot rights cache

Pinning in a chat where the bot may not pin would fail on every single message, so the
pinner checks first.

`BotRights` (`shvatka/tgbot/services/bot_rights.py`, app scoped) maps `chat_id` to the rights
of the bot in that chat and keeps them in memory for 30 minutes. Rights are resolved as:

- owner → can pin;
- administrator → its `can_pin_messages`;
- restricted → its own `can_pin_messages`, while `is_member` (pinning can be allowed to
  everyone but forbidden to a restricted member);
- ordinary member → whatever is allowed to everyone, i.e. the chat's default
  `permissions.can_pin_messages` (a private chat always allows it);
- left or banned → nothing.

Only the ordinary-member case needs a second call (`getChat`); the result is cached either way.
An error while asking telegram is not cached, so it is retried next time instead of blocking
pinning for half an hour.

`MessagePinner.pin` / `.unpin` ask `BotRights.can_pin(chat_id)` and skip the work when the bot
can't pin. Ids of already pinned messages are kept in that case, so they can still be unpinned
once the bot may pin again.

`BotRightsMiddleware` (`dp.my_chat_member.outer_middleware`) refreshes the cache from the
update itself, so promoting or demoting the bot takes effect immediately instead of after the
TTL. It's a middleware rather than a handler so that `my_chat_member` updates stay available to
any handler that may need them later. Since aiogram resolves used update types from handlers
only, `resolve_update_types` adds `my_chat_member` explicitly.

## Tests

- `tests/unit/test_bot_rights.py` - rights per member type, ordinary member against chat
  permissions, caching, expiration, refresh from an update, error handling, the middleware.
- `tests/integration/bot_full/test_pinner.py` - pins all parts, unpins and forgets them,
  categories are independent, nothing is pinned without rights, ids survive a loss of rights,
  telegram errors don't break the flow.
- `tests/integration/bot_full/test_hint_sender.py` - `send_hints` returns caption + all hint parts.

`ruff format --check .`, `ruff check .`, `mypy .` and `pytest tests/unit` pass locally; the
integration suite needs docker, which isn't available in this environment, so it runs in CI.


shvatka/tgbot/handlers/bot_rights.py
```diff
@@ -0,0 +1,29 @@
...
+        "bot membership in chat %s changed to %s",
+        event.chat.id,
+        event.new_chat_member.status,
+    )
+    bot_rights.update(event.chat.id, event.new_chat_member)
+
+
+def setup() -> Router:
+    router = Router(name=__name__)
+    router.my_chat_member.register(bot_rights_changed)
```
bomzheg: we have one more my_chat_member handler, haven't we?
if so, may be do it in middleware?

shvatka/tgbot/services/bot_rights.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def from_member(cls, member: ChatMember) -> "ChatRights":
+        match member:
+            case ChatMemberOwner():
+                return cls(can_pin_messages=True)
+            case ChatMemberAdministrator():
+                return cls(can_pin_messages=bool(member.can_pin_messages))
+            case _:
+                return cls(can_pin_messages=False)
```
bomzheg: oh... in some chats pin can be allowed to everyone but moderator can restrict users for pin...

bomzheg: Right, membership alone isn't enough. Fixed in 38ead46 — rights are now resolved as:

- owner → can pin;
- administrator → its `can_pin_messages`;
- **restricted** → its own `can_pin_messages` (and only while `is_member`), so a member restricted from pinning is respected even when everyone else may pin;
- **ordinary member** → whatever is allowed to everyone, i.e. the chat's default `permissions.can_pin_messages` (private chat → always allowed);
- left / banned → no rights.

`rights_of_member` returns `None` for the ordinary-member case, which is the signal to ask `getChat` — that second call happens only in that case and its result is cached like any other. For the same reason, when a `my_chat_member` update says the bot became an ordinary member, the cache entry is dropped rather than guessed: the next `can_pin` re-resolves it against the chat.

---

shvatka/tgbot/handlers/bot_rights.py
bomzheg: Checked: this was the only `my_chat_member` handler in the repo (`user_join_chat_with_team` is on `chat_member`). But the concern holds anyway — a handler without filters consumes the update, so the next `my_chat_member` handler added to any router would silently never run. Moved to a middleware in 38ead46: `BotRightsMiddleware` on `dp.my_chat_member.outer_middleware`, so the cache is refreshed for every update and the observer stays free for real handlers.

One catch worth flagging: `dp.resolve_used_update_types()` looks at handlers only, so with no handler left `my_chat_member` dropped out of `allowed_updates` and the middleware would never have been reached. `resolve_update_types` now appends it explicitly. Verified on the real bot container: the type is subscribed and the outer middlewares on that observer are `[ContainerMiddleware, BotRightsMiddleware]`, in that order, so `dishka_container` is available.

---

PR #308, opened 2026-07-26
Count key and timer bonuses into game results

bomzheg:
Backend part of #258. UI part: bomzheg/shvatka-ui#155.

`bonus_minutes` was already stored in `event_log` for every key and timer effect, but nothing ever counted it — it was only rendered in bot views. This exposes it so results can show times adjusted by bonuses and penalties.

**No migration needed** — all the data is already there, and permissions are inherited unchanged (`get_game_stat_with_hints` already gates non-complete games to spy-capable orgs).

## Semantics

`bonus_minutes > 0` is a bonus, `< 0` a penalty, so one formula covers both:

```
adjusted = raw - timedelta(minutes=bonus_minutes)
```

Bonuses are attributed to a level via `event_log.level_time_id`. That column is nullable, so there are two fallbacks: resolve the level from the event's timestamp, and failing that keep the event in a `None` bucket that counts only towards the total. A bonus larger than the level's duration is *not* clamped to zero — clamping would break `Σ per-level == total` and would hide a scenario mistake from the organizer.

## Changes

- `GameEventDao.get_game_bonuses_by_teams` — one game-wide query joined to `log_keys` / `timers_log` so the bonus source (key vs. timer) is known. Events with no bonus minutes are filtered out.
- `GameBonusesGetter` protocol composed into `GameStatReader`, so the interactor still takes a single DAO.
- `core/games/results.py` — `resolve_bonus_levels` / `route_bonuses` do the attribution as pure functions; `TeamLevels` gains `bonuses` plus `get_level_bonus` / `get_total_bonus`. The `level_time_id`→level mapping comes from the level times already loaded, so attribution costs no extra query.
- `GET /games/{id}/stat` gains a `bonuses` field, additively — existing clients keep working. Each entry carries the **whole effects** (per review feedback) rather than just the extracted minutes, so new kinds of effect reach clients without an API change:

```json
{
  "level_times": { "42": ["..."] },
  "bonuses": {
    "42": [
      {
        "at": "2026-07-26T21:05:00Z",
        "effects": {
          "id": "...", "bonus_minutes": 5.0, "level_up": false,
          "next_level": null, "hints_": []
        },
        "source": "key",
        "key": "SHКЛЮЧ",
        "level_time_id": 123,
        "level_number": 2
      }
    ]
  }
}
```

(keys of both maps are team ids). Raw times plus a flat list of bonuses, so the client can switch display modes without another request.

- `GameStatWithBonuses` also carries `level_numbers_by_name_id`, the way `CurrentHintsAndKeys` already does, because `Effects.from_core` resolves `next_level` from name_id to `number_in_game`. That costs one extra `add_levels` call in the interactor — deliberate, see the [review thread](https://github.com/bomzheg/Shvatka/pull/308#discussion_r3652410295): deriving the mapping from the level times already loaded would leave `next_level` unresolved when it points at a level no team reached.
- xlsx export gains a bonus block (minutes per team × level, plus a total). Per the issue discussion it carries the raw numbers rather than adjusted times, so they can be computed in Excel; the three existing blocks are untouched, and the block is skipped entirely when a game had no bonuses.
- `Cell.value` accepts `float` (bonus minutes are fractional).
- `AGENTS.md`: new comments and docstrings go in English. The old rule said to follow each file's existing language, which is why this branch first landed Russian ones — see the [review thread](https://github.com/bomzheg/Shvatka/pull/308#discussion_r3652410596).

## Tests

- `tests/unit/domain/bonuses_test.py` — 12 tests: bonus, penalty, both on one level, bonus exceeding the duration, `NULL` / unknown `level_time_id`, event before game start, grouping, source/key survival, and `Σ per-level == total`.
- `tests/integration/api_full/test_game.py` — the `bonuses` payload and its level numbers, that a bonus effect carrying a hint round-trips with the hint intact, that a no-bonus event is excluded, and that a game with no bonuses returns `{}`.

`ruff format --check`, `ruff check` and `mypy` are clean; `pytest tests/unit` passes (213). Integration tests need Docker, which isn't available in this environment — CI covers those, and is green.

## Open question

The review asked whether the UI should filter, which may have meant returning *all* events rather than only those with bonus minutes. I kept the server-side filter to avoid putting every level-up key event into a payload that gets polled during play — happy to drop it, it's a small change. See the thread linked above.

## Not in this PR

The chart (`game_chart.part`, `infrastructure/picture`), bonuses in live standings of an unfinished game, `effects` on `GET /games/{id}/keys`, and bot-side rendering.

### DIFF
--- a/shvatka/api/models/responses.py
+++ b/shvatka/api/models/responses.py
@@ -6,7 +6,14 @@ from uuid import UUID
 
 from adaptix import Retort
 
-from shvatka.core.games.dto import CurrentHintsAndKeys, MyRole, Event
+from shvatka.core.games.dto import (
+    CurrentHintsAndKeys,
+    MyRole,
+    Event,
+    BonusSource,
+    GameStatWithBonuses,
+    BonusEvent as CoreBonusEvent,
+)
 from shvatka.core.models import dto, enums
 from shvatka.core.notifications.dto import (
     Notification as NotificationDto,
@@ -586,18 +593,48 @@ class LevelTime:
         )
 
 
+@dataclass(kw_only=True, frozen=True, slots=True)
+class BonusEvent:
+    """Бонус (minutes > 0) или штраф (minutes < 0), полученный командой."""
+
+    at: datetime
+    minutes: float
+    source: BonusSource
+    key: str | None
+    level_time_id: int | None
+    level_number: int | None
+    """Уровень, на котором получен. None — определить не удалось, учитывать только в итоге."""
+
+    @classmethod
+    def from_core(cls, core: CoreBonusEvent) -> "BonusEvent":
+        return cls(
+            at=core.at,
+            minutes=core.minutes,
+            source=core.source,
+            key=core.key,
+            level_time_id=core.level_time_id,
+            level_number=core.level_number,
+        )
+
+
 @dataclass
 class GameStat:
     level_times: dict[int, list[LevelTime]]
+    bonuses: dict[int, list[BonusEvent]]
+    """{team_id: [...]} — только команды, у которых бонусы есть."""
 
     @classmethod
-    def from_core(cls, core: dto.GameStatWithHints | None):
+    def from_core(cls, core: GameStatWithBonuses | None):
         if core is None:
             return None
         return cls(
             level_times={
                 team.id: [LevelTime.from_core(lt) for lt in lts]
                 for team, lts in core.level_times.items()
+            },
+            bonuses={
+                team_id: [BonusEvent.from_core(bonus) for bonus in bonuses]
+                for team_id, bonuses in core.bonuses.items()
             },
         )
 
--- a/shvatka/api/models/responses.py
+++ b/shvatka/api/models/responses.py
@@ -595,21 +595,27 @@ class LevelTime:
 
 @dataclass(kw_only=True, frozen=True, slots=True)
 class BonusEvent:
-    """Бонус (minutes > 0) или штраф (minutes < 0), полученный командой."""
+    """An event that changed a team's time, with the whole effects that caused it.
+
+    The bonus itself is ``effects.bonus_minutes``: positive is a bonus, negative
+    a penalty. Only events that carry bonus minutes are returned.
+    """
 
     at: datetime
-    minutes: float
+    effects: Effects
     source: BonusSource
     key: str | None
     level_time_id: int | None
     level_number: int | None
-    """Уровень, на котором получен. None — определить не удалось, учитывать только в итоге."""
+    """Level it was earned on. None when unresolved — then count it in the total only."""
 
     @classmethod
-    def from_core(cls, core: CoreBonusEvent) -> "BonusEvent":
+    def from_core(
+        cls, core: CoreBonusEvent, level_numbers_by_name_id: Mapping[str, int]
+    ) -> "BonusEvent":
         return cls(
             at=core.at,
-            minutes=core.minutes,
+            effects=Effects.from_core(core.effects, level_numbers_by_name_id),
             source=core.source,
             key=core.key,
             level_time_id=core.level_time_id,
@@ -621,7 +627,7 @@ class BonusEvent:
 class GameStat:
     level_times: dict[int, list[LevelTime]]
     bonuses: dict[int, list[BonusEvent]]
-    """{team_id: [...]} — только команды, у которых бонусы есть."""
+    """{team_id: [...]} — only teams that actually have bonuses."""
 
     @classmethod
     def from_core(cls, core: GameStatWithBonuses | None):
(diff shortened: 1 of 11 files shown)


shvatka/api/models/responses.py
```diff
@@ -586,19 +593,49 @@ def from_core(cls, core: dto.LevelTimeOnGame | None):
...
+            key=core.key,
+            level_time_id=core.level_time_id,
+            level_number=core.level_number,
+        )
+
+
 @dataclass
 class GameStat:
     level_times: dict[int, list[LevelTime]]
+    bonuses: dict[int, list[BonusEvent]]
```
bomzheg, line 661: may be whole effects? it's more extendable. ui can filter it itself, i guess

shvatka/core/games/dto.py
```diff
@@ -16,6 +17,58 @@ class Event:
...
+        """Сколько времени бонус снимает с результата (штраф — отрицательный)."""
+        return timedelta(minutes=self.minutes)
+
+
+@dataclass(kw_only=True, frozen=True, slots=True)
+class GameStatWithBonuses:
+    """Статистика игры вместе с бонусами и штрафами команд.
+
+    Скорректированные времена здесь не считаются: отдаём исходные времена и
+    сами бонусы, чтобы клиент мог переключать режимы отображения без запросов.
```
bomzheg: make comments in English

shvatka/api/models/responses.py
bomzheg: Done in f050e96 — `BonusEvent` now carries the whole `effects` instead of `minutes`, reusing the existing `responses.Effects` model, so it's the same shape the UI already handles in `KeyWithEffects` and `GameEvent`.

Two consequences worth flagging:

**`next_level` resolution needs the name_id mapping.** `Effects.from_core` resolves `next_level` from name_id to `number_in_game`, so `GameStatWithBonuses` now carries `level_numbers_by_name_id`, the same way `CurrentHintsAndKeys` already does. That costs one extra `add_levels` call in the interactor. I could have derived the mapping from the `LevelTimeOnGame.name_id` values already in the stat, but that only covers levels some team actually reached — a `next_level` pointing at an unvisited level would come back unresolved in a routed game. The extra query buys correctness there.

**I kept the server-side filter to events that carry bonus minutes.** So it's still "the bonuses", not "all events" — a pure level-up key event doesn't appear. My reasoning: dropping the filter would put every key event in the results payload, and on a long game with many teams that's a lot of data for an endpoint that polls during play. `minutes`/`td` remain on the core DTO as derived properties, so the domain arithmetic didn't change.

If you did mean the wider thing — return *all* events and let the UI decide what's a bonus — say so and I'll drop the filter. That's a one-line change in the DAO plus renaming the field from `bonuses`; the UI already filters by `bonus_minutes !== 0` when it builds its per-level maps, so it wouldn't need much either.

Added an integration assertion that a bonus effect carrying a hint round-trips with the hint intact, so the extensibility is actually covered.

---

shvatka/core/games/dto.py
bomzheg: Done in f050e96 — translated every comment and docstring this branch added, in both repos. User-facing strings stay Russian: the bot texts, the Excel headers (`Бонусы, мин` / `Итого`), and the UI labels and tooltips.

I also updated the convention in `AGENTS.md`, since it actively pointed the other way — it said "some docstrings/comments are in Russian — that's expected; keep the existing language of the file you're editing," which is what I followed. It now says new comments go in English while user-facing strings stay Russian, and notes there's no need to translate existing ones when touching a file. Added the equivalent line to the UI repo's `AGENTS.md`, which only covered user-facing strings.

---

shvatka/api/models/responses.py
bomzheg: You're right, and it's a stronger argument than "not secret": `responses.LevelTime` in this very response **already carries `name_id`**, so resolving it in `effects` hid nothing while costing a query.

Done in 69545a8 (test fix in f3617ae). Two response models now:

- `Effects` — `next_level: int | None`, resolved from name_id. For endpoints a playing team can read, where a name_id would leak scenario structure. Unchanged, still used by `KeyWithEffects` and `GameEvent`.
- `EffectsWithNameId` — `next_level: str | None`, passed through as stored. Used by the results endpoint.

So `GameStatWithBonuses` no longer carries `level_numbers_by_name_id` and the interactor no longer calls `add_levels` — the extra query I flagged in my previous reply is gone, and so is the routed-game edge case I was worried about, since nothing needs resolving any more.

No UI change needed: `EffectLike.next_level` is already typed `number | string | null` and `scenario_graph.model.ts` already resolves either form.

Added an assertion that a bonus effect routing to another level comes back with that level's name_id rather than its number.

One note on CI: my first push of this failed the test job, and it was my own test bug rather than the change — I wrote `finished_game.levels[2]` when `complex_scn` has two levels, so it raised `IndexError`. Fixed in f3617ae by routing to `levels[0]`, which is a more meaningful case anyway (a route-back from the level the event fires on). Green now.

---

PR #309, opened 2026-07-26
Generate player username from transliterated tg names

bomzheg:
## What

A new telegram player got `id{id}` as a shvatka username whenever telegram gave
no username. Now the username is built from the transliterated telegram first
and last name, and the id-based one is kept only for players who have no name at
all.

Username sources for a new player, in order:

1. telegram username (as before);
2. forum name (as before);
3. transliterated telegram first + last name — **new**, e.g. `Гарри Поттер` →
   `garri_potter`, `Гарри` → `garri`;
4. `id{id}` / `id{id}_{n}` (as before).

## How

- `shvatka/core/utils/username.py` — new pure-domain module:
  - `transliterate()` maps cyrillic (ru/ua/by letters) char by char, drops
    diacritics of latin-based scripts (`Renée` → `renee`) and squashes
    everything else (punctuation, emoji, hieroglyphs) into a single underscore;
  - `username_from_names()` joins the two names and validates the result with
    the existing `validate_new_username`, so anything too short or with no
    latin-able characters left yields `None` and the id fallback applies;
  - `numbered_username()` builds the `_1`, `_2`… variants without exceeding the
    50-char limit.
- `PlayerDao.get_free_and_associated_username` is split into
  `find_associated_username` (name-based, may return `None`) and
  `_free_id_username` (the old id fallback), so both player creation and the
  updater script share one generation rule.
- Occupied name-based candidates get a numeric suffix (`garri_potter_1`).

## Updating already saved players

New entry point `shvatka-player-username-updater`
(`shvatka/infrastructure/player_username_updater.py`) selects players whose
username is exactly `id{id}`, regenerates it from their linked telegram/forum
identity and commits, logging every rename. Players with no usable name keep
their id-based username, so the script is safe to re-run.

## Tests

- `tests/unit/domain/username_generation_test.py` — transliteration, name
  joining, length limits, unusable names.
- `tests/integration/test_player_username.py` — username by tg username, by
  transliterated names, by first name only, collision suffix, id fallback, and
  the updater renaming (and not renaming) saved players.

`ruff format`, `ruff check`, `mypy` and `pytest tests/unit` pass locally;
integration tests need Docker, so they run in CI.

## Note

`is_username_occupied` matches with `ILIKE` without escaping, so `_` in a
candidate acts as a wildcard and a name-based username can be reported occupied
by a similar-looking one (`garri_potter` vs `garrixpotter`). That only ever adds
a numeric suffix, never breaks uniqueness, so this PR leaves the pre-existing
behaviour (shared with login lookup) untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_013TTaDFTX1tNf6f5PCXFbrg)_

### DIFF
--- a/shvatka/infrastructure/player_username_updater.py
+++ b/shvatka/infrastructure/player_username_updater.py
@@ -1,0 +1,49 @@
+import asyncio
+import logging
+
+from dishka import make_async_container
+
+from shvatka.common import setup_logging
+from shvatka.common.config.parser.paths import common_get_paths
+from shvatka.infrastructure.db.dao.holder import HolderDao
+from shvatka.infrastructure.di import get_providers
+from shvatka.infrastructure.di.infra import get_infra_only_providers
+
+logger = logging.getLogger(__name__)
+
+
+async def main():
+    paths = common_get_paths("INFRA_PATH")
+
+    setup_logging(paths)
+    dishka = make_async_container(
+        *get_providers("INFRA_PATH"),
+        *get_infra_only_providers(),
+    )
+    try:
+        dao = await dishka.get(HolderDao)
+        await renew_id_usernames(dao)
+    finally:
+        await dishka.close()
+
+
+async def renew_id_usernames(dao: HolderDao) -> list[tuple[str, str]]:
+    """Give players with an `id{id}` username a name-based one instead.
+
+    Players are renamed only if their telegram or forum identity provides a free
+    username, so those without any name keep the id-based one.
+    """
+    renamed = await dao.player.renew_id_usernames()
+    for old, new in renamed:
+        logger.info("username %s renewed to %s", old, new)
+    await dao.commit()
+    logger.info("renewed %s usernames", len(renamed))
+    return renamed
+
+
+def run():
+    asyncio.run(main())
+
+
+if __name__ == "__main__":
+    run()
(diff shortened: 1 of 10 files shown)


shvatka/infrastructure/player_username_updater.py
```diff
@@ -0,0 +1,49 @@
...
+from shvatka.infrastructure.di.infra import get_infra_only_providers
+
+logger = logging.getLogger(__name__)
+
+
+async def main():
+    paths = common_get_paths("INFRA_PATH")
+
+    setup_logging(paths)
+    dishka = make_async_container(
```
bomzheg, line 19: i guess this create container with APP scope, we need REQUEST scope for get dao
probably same is in other scripts. pls fix

bomzheg: Fixed in 8ecc715 — the dao (and file storage / gateway) are now resolved inside `async with dishka() as request_dishka`, like `stored_scn_checker` already does.

Same problem was in `file_id_updater`, `empty_file_repairer`, `crawler/teams/loader` and `crawler/game_scn/loader/load_scns` — all of them asked the APP container for `HolderDao` / `FileStorage` / `FileGateway`, which are `Scope.REQUEST`; fixed there too. APP-scope deps (`TgBotConfig`, `Factory`, `Retort`) still come from the app container.

---

PR #310, opened 2026-07-26
Let admins change a player's username

bomzheg:
The admin panel could already change a player's email and relink their telegram account. This adds the third identity field: the username itself.

## Changes

- `AdminSetPlayerUsernameInteractor` (`core/players/admin_interactors.py`) — authorises through `identity.get_superuser()`, logs the acting admin, validates the new username with `validate_new_username` (same rules as the self-service change) and delegates to the existing `set_player_username`, which rejects an occupied one.
- `AdminUsernameSetter` protocol + `AdminUsernameSetterImpl` adapter, wired into `AdminProvider`.
- `PUT /admin/players/{id}/username` returning the refreshed `PlayerWithIdentities`, so the UI can re-render the whole card. Invalid username → 422, occupied → 409 — matching the mapping the email-register route already uses.

Renaming a player to their own current username is a no-op success (`is_same_username` short-circuits the occupied check), which keeps the UI's "save" button harmless when nothing was edited.

## Second commit: a stale user in the session

The first push turned `tests/integration/test_player_username.py::test_renew_id_usernames` red, and it was not the new endpoint's doing — the test asserts that a player carrying an `id{id}` username gets renamed from their telegram names, and it got no rename at all.

Root cause: `UserDao.upsert_user` runs `INSERT … ON CONFLICT DO UPDATE … RETURNING User`, and sessions are created with `expire_on_commit=False`. When the user being upserted is already in the session's identity map, the ORM hands back that mapped instance without applying the returned row, so the upsert's own return value — and every relationship pointing at that user — keeps the pre-upsert names. `renew_id_usernames` then reads `player.user.first_name` as `None`, finds no name to build a username from, and renames nobody.

The test only ever passed because the mapped instance is weakly referenced: when it happened to be garbage-collected before the next read, the relationship reloaded from the database and saw the new names. That is why master is green and why this branch failed twice — the latent bug predates it.

Fix: pass `populate_existing=True` on that execute, so the returned row wins over what the session already holds. `tests/integration/test_save_user.py::test_upsert_user_refreshes_already_loaded_user` pins the mapped instance in the session and asserts both the returned dto and the mapped user carry the new names, so it fails deterministically without the fix rather than depending on the garbage collector.

## Tests

Integration tests in `tests/integration/api_full/test_admin.py`: successful rename, invalid username, occupied username, renaming to the current username, and the non-superuser 403. Plus the upsert regression test above.

Verified locally against a real Postgres: `ruff format --check`, `ruff check`, `mypy` clean, and the full suite green (498 passed).

### DIFF
--- a/shvatka/api/models/req.py
+++ b/shvatka/api/models/req.py
@@ -104,6 +104,11 @@ class AdminChangeEmail:
 
 
 @dataclass
+class AdminChangeUsername:
+    username: str
+
+
+@dataclass
 class AdminChangeTg:
     tg_id: int
     username: str | None = None
--- a/shvatka/api/routes/admin.py
+++ b/shvatka/api/routes/admin.py
@@ -23,6 +23,7 @@ from shvatka.core.players.admin_interactors import (
     AdminMergePlayersInteractor,
     AdminSearchPlayersInteractor,
     AdminSetPlayerEmailInteractor,
+    AdminSetPlayerUsernameInteractor,
 )
 from shvatka.core.services.one_time_link import GenerateOneTimeLoginLinkForPlayerInteractor
 from shvatka.core.services.scenario.files import get_file_metas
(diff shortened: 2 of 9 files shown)


dump.rdb
bomzheg, line 1: do not commit it

bomzheg: Sorry — that was a stray artifact. I ran a local Redis to reproduce the CI failure without Docker, it wrote its snapshot into the repo root, and my `git add -A` swept it in.

Removed: I amended the commit rather than adding a delete on top, so it is out of the branch history entirely (force-pushed, `5ade010`). Redis now runs with its `dir` outside the repo and snapshots off, so it will not come back.

---

PR #312, opened 2026-07-27
Move API routes next to their models, plumbing into api/app

bomzheg:
Follow-up to #311, which split the models by subdomain. This finishes the job: the endpoints move next to the models they speak, and the framework plumbing moves out of the way.

## What

```
shvatka/api/
  __main__.py        # entrypoints stay at the top
  main_factory.py    #   create_app / get_paths
  password_hash.py   #   shvatka-password script
  app/               # framework plumbing — nothing subdomain-specific
    router.py        #   aggregates every subdomain router (was routes/__init__.py)
    error_handler.py #   SHError -> HTTP response
    config/          #   ApiConfig models + parsers
    dependencies/    #   dishka providers, ApiIdentityProvider, AuthProperties
    middlewares/     #   logging, CORS
    utils/           #   cookie auth, web views/notifiers, push sender
  shared/            # models used by more than one subdomain
  auth/  players/  teams/  games/  waivers/  files/
  notifications/  action_requests/  search/  push/  admin/  version/
      └─ routes.py + requests.py + responses.py
```

Route module → subdomain, for review:

| was | now |
| --- | --- |
| `routes/user.py` | `players/routes.py` |
| `routes/team.py` | `teams/routes.py` |
| `routes/game.py` | `games/routes.py` |
| `routes/cdn.py` | `files/routes.py` |
| `routes/requests.py` | `action_requests/routes.py` |
| `routes/__init__.py` | `app/router.py` |
| `routes/error_handler.py` | `app/error_handler.py` |

Subdomains keep the short top-level paths since they're what gets edited; the plumbing goes one level deeper. `app/` holds only what every subdomain shares — if something there knows about one subdomain, it belongs in that subdomain.

## Diff shape

**95 files, +208/−167.** Almost every file is a git-detected rename with only its import lines touched; `app/router.py` is the one file rewritten (it now imports each subdomain's `routes` module instead of sibling modules). Reviewing the non-rename hunks is enough.

## Behavior

None intended — pure reorganisation. Verified mechanically: the **generated OpenAPI schema is byte-identical** to master, diffed by building the router under both trees.

## Checks

- `ruff check .` — clean
- `ruff format --check .` — clean
- `mypy .` — no issues in 750 source files
- `pytest tests/unit` — 235 passed

Integration tests (`tests/integration/api_full/`) had their imports updated but need testcontainers, so they run in CI.

## Docs

`AGENTS.md` gains an "API layout (subdomain packages)" section with the placement rules, and its stale `api/routes/…` / `api/dependencies/…` path references are updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_0124313JLXGoFFV4bfkYCyPt)_


shvatka/api/app/router.py
```diff
@@ -0,0 +1,31 @@
+from fastapi import APIRouter
+
+from shvatka.api.action_requests import routes as action_requests
+from shvatka.api.admin import routes as admin
```
bomzheg, line 4: we can export setup admin.__init__ so we can have this without rename import

bomzheg: Tried it — `from shvatka.api.<sub>.routes import setup` + `__all__` in each `__init__.py`, and `from shvatka.api import (action_requests, admin, auth, ...)` here. It works for 11 of the 12, but `auth` closes a real import cycle and the app stops booting entirely:

```
shvatka/api/app/router.py:3            from shvatka.api import (...)
shvatka/api/action_requests/__init__.py:1  from shvatka.api.action_requests.routes import setup
shvatka/api/action_requests/routes.py:6    from shvatka.api.app.dependencies.auth import ApiIdentityProvider
shvatka/api/app/dependencies/__init__.py:5 from shvatka.api.app.dependencies.auth import AuthProvider
shvatka/api/app/dependencies/auth.py:17    from shvatka.api.auth.requests import UserTgAuth
shvatka/api/auth/__init__.py:1             from shvatka.api.auth.routes import setup
shvatka/api/auth/routes.py:39              from shvatka.api.app.dependencies.auth import AuthProperties, ...
ImportError: cannot import name 'AuthProperties' from partially initialized module
'shvatka.api.app.dependencies.auth' (most likely due to a circular import)
```

The auth plumbing imports auth's own models — `dependencies/auth.py` needs `UserTgAuth` and `Token`, `utils/cookie_auth.py` needs `Token`. Making `auth/__init__.py` pull in `auth/routes.py` turns that into a loop. It's not order-dependent luck either: it fails whenever anything reaches `app.dependencies.auth` before `shvatka.api.auth`, which is most entry points (including plain `import shvatka.api.main_factory` and the test conftest).

So I've left the aliased imports in for now. Two ways to get what you're asking for, if you want one:

1. **Move `Token` out of `auth/responses.py` into `app/`** (next to `AuthProperties` / the cookie helper) and import `UserTgAuth` under `TYPE_CHECKING` in `dependencies/auth.py`. That breaks the cycle at its root and lets all 12 packages re-export `setup`. Defensible on its own merits: no route actually returns `Token` — `auth/routes.py` returns `{"ok": True}` and sets a cookie — so it's a mechanism type, not an endpoint response.
2. **Re-export from the 11 and leave `auth/__init__.py` empty**, with a comment saying why. Less churn, but a mixed style here and a trap for whoever adds the next plumbing→subdomain-model import.

I'd go with 1. Say the word and I'll push it.

---

PR #314, opened 2026-07-27
Релиз игры: an optional pre-game release (#157)

bomzheg:
Closes part of #157 (engine side).

A game's **release** ("Релиз игры") is the promo published before it. It leads with a **banner** — a wide title picture with a caption — followed by the rest: some text about the theme, a map. The part after the banner is a plain `list[AnyHint]`, so the whole existing hint machinery (message parser, hint sender, web renderers, hint editors) works for it unchanged. Both halves are optional, and so is the release: nothing in the game flow requires one.

## Why the banner is separate

The site shows a release above the header, where the countdown lives — and an arbitrary list of hint parts does not fit there. The banner is the one part that does: it gets its own field and its own column, so the header can read it alone, without the rest. Roughly 1280×250—1280×550 in practice, though nothing enforces that. A release may be just a banner, or have none at all.

## Lifecycle

Writing the release and announcing it are separate. The author writes it whenever; the engine decides when it reaches the announcements channel:

| release written… | what happens |
|---|---|
| before the waivers (`underconstruction` / `ready`) | stored; posted when the game turns to `getting_waivers` |
| while the game is already `getting_waivers` | posted at once |
| after the game started (and later) | stored only — its audience is already playing |
| edited while it stands in the channel | those messages are **edited in place** |
| deleted | the channel messages are deleted too |

The waivers hook lives in `ChangeGameStatusInteractor`, which both the bot's "Начать сбор вейверов" and the web's status change already go through.

**Editing window:** open up to and including a `finished` game — a release is promo, not part of the play. A `complete` game's release is history: only an admin may still change it.

## Editing in the channel

`GameBotReleasePublisher` posts one message per part — the banner first — and edits those messages afterwards, so the channel keeps a single announcement per game: `edit_message_text` for text, `edit_message_media` for photo/video/animation/audio/document. When telegram can't do it — a part added or removed, a kind of content swapped, or a file it has never seen — the old messages are deleted and the release is posted anew.

## Model

- `games.release` (jsonb, the body) + `games.release_banner` (jsonb, the leading photo) + `games.release_post` (jsonb: chat id and message ids), with migrations
- `dto.GameRelease(game_id, banner, hints, post)` — a separate aggregate, so no existing `dto.Game` call site changed. `parts` gives the release in the order it is shown
- `dto.ReleasePost` is deliberately not called a *publication*: `context.md` reserves that word for a finished game's results
- writes live in the games table's own DAO; reading selects only the release columns, so it never parks a partially-loaded `Game` in the identity map

## API

| method | route | who |
|---|---|---|
| GET | `/games/{id}/release` | anyone (`null` when the game has none) |
| PUT | `/games/my/{id}/release` | author, or an admin for a complete game |
| DELETE | `/games/my/{id}/release` | same |

`PUT` takes `banner` alongside `hints`. `GameFileReaderInteractor` serves files a release references without auth, so the banner renders for guests too.

## Bot

`📢Релиз` in the game menu → `📝Собрать релиз` → **the banner** (send a photo, with a caption if you like, or skip it) → **the rest** (any messages, collected in `dialog_data`) → `👁Предпросмотр`, which sends the whole thing back exactly as the channel will render it → `✅Сохранить`. The confirm screen says what saving will do — go out now, wait for the waivers, or stay on the site.

The menu button used to be gated on `can_be_edited`, which freezes along with the scenario at `getting_waivers`. That contradicted the editing window above and hid the release outright for a started game, so the guard is gone — the interactor already enforces the real rule.

## DI note

Announcing needs a bot, and waivers can start from the site, so `HintSender` and its content resolver moved from the bot-only providers into the shared ones; the publisher sits in its own provider so the integration tests announce into a mock. `AGENTS.md` records the rule.

## Glossary

`context.md` gains **Релиз**, **Баннер** and **Пост релиза**, the statuses section says how far a release may be edited and when it goes out, and *анонс* joins the words we don't write. The same terms are added to the front-end's `context.md` in bomzheg/shvatka-ui#157.

## Tests

- `tests/unit/services/game_release_interactors_test.py` — each branch of the lifecycle, post vs edit vs delete, the editing window and the admin rule, a banner-only release and dropping a banner
- `tests/integration/api_full/test_game_release.py` — the endpoints end to end, including a banner uploaded through the cdn and read back by a guest, and a complete game refusing its author while accepting an admin

`ruff format`, `ruff check`, `mypy` and `pytest tests/unit` (301) pass locally on this base; the integration suite needs Docker, so it runs in CI.


shvatka/infrastructure/di/hints.py
```diff
@@ -0,0 +1,68 @@
+"""Sending hints to telegram — shared by the bot and the api.
```
bomzheg: please don't share. when pin we already have same pattern. all tgbot show should be tgbot responsibility. also storing message_ids, chat_ids. it's part of view - so view should care about it.

interactor can call view (for separate process for ui and bot) or submit to event bus)

bomzheg: Done in f56788e — followed the pinner pattern throughout.

**Nothing shared.** `infrastructure/di/hints.py` is deleted; `HintSender` and `HintContentResolver` are back in the bot-only `GameToolsProvider`, exactly as on master. The publisher joins them in `tgbot`.

**The interactor calls a view.** `GameReleasePublisher` is a Protocol in `core/views/`, bound per container like `GameView` and `GameLogWriter` already are — the real one in `tgbot/main_factory.py`, `WebGameReleasePublisher` (does nothing: the site reads the release as it is) in `api/app/dependencies/`, and the existing no-op for the infra scripts. So a release still reaches the channel when the waivers are started from the site, without the bot's tools in the api container.

**Message ids are the view's.** `ReleasePostDao` keeps them in redis next to `PinnedMessageDao`; `games.release_post` and `dto.ReleasePost` are gone, with a migration dropping the column. Since only the view knows whether it is showing anything, the protocol became `publish` (show it — post or bring up to date) and `update` (refresh only what is already up), instead of handing a post back and forth.

That leaves the interactor with the one thing that is actually domain: whether the audience should see the release now, i.e. whether the game is collecting waivers.

Two consequences worth flagging:

- `is_published` is gone from `GET /games/{id}/release` — the engine can no longer answer it. The web editor's copy now comes from the game's status alone (bomzheg/shvatka-ui#157, 231bc3c).
- The posted messages now live in redis rather than postgres, so losing redis loses the link to them: a later edit would post the release again instead of editing. Tell me if you'd rather that state were durable — it just needs a different dao behind the same view.

I also rewrote the AGENTS.md paragraph I had added, since it advised the opposite of this.

---

AGENTS.md
```diff
@@ -142,6 +142,26 @@ async def get_game_stat(
...
+that needs the chat to change takes a **view Protocol** from `core/views/` and
+each container binds its own implementation: the bot one in
+`tgbot/main_factory.py`, a web one in `api/app/dependencies/`, a no-op in
+`infrastructure/di/infra.py`. (Where an event fits better than a call, submit
+one to the `Bus` instead.) That way the same use case reaches the channel from
+the site, without `HintSender` leaking into every container.
+
+**What a view showed is the view's to remember**, not the game's. Chat ids and
+message ids never reach the domain or the games table: the view keeps them
+beside the bot's other bookkeeping in redis — see `MessagePinner` with
```
bomzheg: it ok, to save it into pg, just do not leak it into the core domain entity.
it's ok to load it to another dto, or without dto. f.e. action request bot messages. it stored without dto (but we can have some dto for it. just not for core entity

so let's restore db column. just do not load it to domain entity

bomzheg: Restored in b8c255c, following `action_requests.bot_messages`.

`games.release_post` is back exactly as it was, so the schema is unchanged again — the redis dao and the drop migration are both gone, and the two release migrations are the only ones left. It is read and written through `GameDao.get_release_post` / `save_release_post` / `clear_release_post`, which return and take plain values (`tuple[int, list[int]]`), are called only by `GameBotReleasePublisher`, and never appear in `to_dto`. No chat or message id reaches `dto.Game` or `dto.GameRelease`.

One thing that fell out of the move: `delete_release` no longer nulls `release_post`. It used to clear it in the same statement, which — now that taking the messages down happens after that commit — would have left the view unable to find them. The view clears the column itself once the messages are gone.

Durability worry from my previous comment is moot, since it is back in postgres.

---

shvatka/infrastructure/db/dao/rdb/game.py
```diff
@@ -276,6 +277,68 @@ async def set_keys_url(self, game: dto.Game, url: str):
...
+    async def delete_release(self, game: dto.Game) -> None:
+        # release_post is not cleared here: it is the view's, and the view still
+        # needs it to take the messages down after this commit
+        await self.session.execute(
+            update(models.Game)
+            .where(models.Game.id == game.id)
+            .values(release=None, release_banner=None)
+        )
+
+    async def get_release_post(self, game_id: int) -> tuple[int, list[int]] | None:
```
bomzheg: tuple[int, list[int]] as in action request? lets add shared dataclass for it. we can add it in core, just do not use it in things out of tgbot

shvatka/tgbot/views/game_release.py
```diff
@@ -0,0 +1,155 @@
...
+
+    async def take_down(self, game_id: int, chat_id: int, message_ids: list[int]) -> None:
+        for message_id in message_ids:
+            await self.delete_one(chat_id, message_id)
+        await self.dao.clear_release_post(game_id)
+        await self.dao.commit()
+
+    async def delete_one(self, chat_id: int, message_id: int) -> None:
+        try:
+            await self.bot.delete_message(chat_id=chat_id, message_id=message_id)
```
bomzheg: may be use our function total_delete_msg (I don't really remember name) to delete?

bomzheg: `total_remove_msg` — used in 11c9ed7.

Better than what was here, too: it falls back to editing the message to a struck-through "(удалено)" when telegram won't delete it any more, so the channel never keeps a release that is supposed to be gone. My version just logged and moved on.

---

shvatka/infrastructure/db/dao/rdb/game.py
bomzheg: Added as `dto.BotMessage(chat_id, message_id)` in 11c9ed7, and it now serves both places — `get_bot_messages` returns `Sequence[BotMessage]` too, so the action-request tuple is gone as well.

Taking the pair as the shared unit rather than a release-shaped `(chat, [messages])` means a release post is simply a list of them, which also makes `games.release_post` the same jsonb shape as `action_requests.bot_messages` — `[{"chat_id": …, "message_id": …}, …]`.

It sits in `core/models/dto/bot_message.py`, and stays out of every entity as you asked: only the daos that keep these and the views that posted them ever hold one. The glossary gains a **Bot message** row saying exactly that.

Note the stored shape of `release_post` changed from `{"chat_id", "message_ids"}` to that list. The column is unmerged, so no migration — but a release already posted from a test deployment will read as "nothing posted" and be posted afresh on the next edit.

---

shvatka/infrastructure/di/interactors.py
```diff
@@ -287,9 +303,20 @@ def change_start_at(
...
+        dao: HolderDao,
+        game_log: GameLogWriter,
+        release_dao: GameReleaseGetter,
+        release_publisher: GameReleasePublisher,
+    ) -> ChangeGameStatusInteractor:
         return ChangeGameStatusInteractor(
-            getter=dao.game, waiver_starter=dao.game, completer=dao.game, game_log=game_log
+            getter=dao.game,
+            waiver_starter=dao.game,
+            completer=dao.game,
```
bomzheg: how about merge it in one complex dao and pass it to di? then this interactor we can provide by the function call:
change_status = provide(ChangeGameStatusInteractor)

shvatka/infrastructure/di/interactors.py
```diff
@@ -331,6 +358,28 @@ def remove_org(self, dao: HolderDao) -> RemoveGameOrgInteractor:
...
+    def game_release_reader(self, dao: HolderDao) -> GameReleaseReader:
+        return GameReleaseReaderImpl(dao=dao)
+
+    @provide
+    def game_release_editor(self, dao: HolderDao) -> GameReleaseEditor:
+        return GameReleaseEditorImpl(dao=dao)
+
+    @provide
+    def game_release_getter(self, dao: HolderDao) -> GameReleaseGetter:
+        return GameReleaseReaderImpl(dao=dao)
```
bomzheg: use provide() function for simple cases

shvatka/infrastructure/di/interactors.py
bomzheg: Done in 88b3c82 — `change_status = provide(ChangeGameStatusInteractor)`.

All four collaborators were the same object anyway: `getter`, `waiver_starter`, `completer` and (since the announcer went away) the release reader were each `dao.game`. They are now one `GameStatusChanger` next to `GameCompleter` in `interfaces/dal/complex.py`:

```python
class GameStatusChanger(
    GameByIdGetter, WaiverStarter, GameCompleter, GameReleaseGetter, Protocol
): ...
```

The release getter belongs in it because starting the waivers is exactly what puts the release in front of people. `game_release_getter` is gone with the factory — nothing needs a bare `GameReleaseGetter` any more.

---

shvatka/infrastructure/di/interactors.py
bomzheg: Partly, in 88b3c82.

`game_release_getter` — the one at this line — is deleted outright: with `GameStatusChanger` covering the status change, nothing asks for a bare `GameReleaseGetter`. The three release interactors were already `provide(...)`, and `change_status` becomes one too.

The two remaining factories I tried to convert and had to put back. `provide(GameReleaseReaderImpl, provides=GameReleaseReader)` fails at container build:

```
dishka.provider.exceptions.UndefinedTypeAnalysisError: Failed to analyze
`...complex.game.GameReleaseReaderImpl`  →  NameError: name 'HolderDao' is not defined
```

The impls annotate `dao: "HolderDao"` and import `HolderDao` only under `TYPE_CHECKING`, so there is nothing for dishka to resolve at runtime. That is why every complex dao in this file is written out longhand — `game_scenario_editor`, `admin_game_scenario_editor` and the rest. I left a comment saying so, so nobody spends the same ten minutes on it.

Happy to make them `provide()`-able instead if you'd rather: it needs the impls to import `HolderDao` for real, which is a circular-import question in `dao/complex/game.py` rather than a DI one.

---

PR #335, opened 2026-08-08
Parse config with dature instead of dataclass_factory/adaptix

bomzheg:
Config loading moves to [dature](https://github.com/reagento/dature). Only the loader functions changed — every config model stays a plain `@dataclass` with no dature decorators and no loading metadata on it.

## How it works now

`shvatka/common/config/parser/config_source.py` builds a dature source over `config.yml`:

```python
def config_source(paths, prefix=None, field_mapping=None, type_loaders=None) -> dature.Yaml11Source
```

A loader passes the dot-separated `prefix` of the section it owns and hands the source to `dature.load(..., schema=TheModel)`:

```python
def load_db_config(paths: Paths) -> DBConfig:
    return dature.load(config_source(paths, prefix="db"), schema=DBConfigProperties)
```

Every loader takes `Paths` now instead of a raw dict plus a `Factory`, so `read_config()` and the `dcf` threading through `load_config` are gone. What disappeared with them:

- manual key digging — `dct["secret-key"]`, `config_dct.get("mail")`, `config_dct["api"].get("context-path", "")`
- per-field coercion — `bool(...)`, `int(...)`, `timedelta(minutes=int(...))`, `StorageType[dct["type"]]`
- the `{**config_dct["bot"], "superusers": ...}` dict splice; `superusers` now reaches `BotConfig` via `field_mapping={F[BotConfig].superusers: Absolute("superusers")}`, which reads the root key while the rest of the section comes from `bot`
- `load_storage_config`'s manual `if type_ == redis` branch — `StorageConfig` loads whole, and `create_storage` already raises a clear error when a redis storage has no redis config

Two sections of `config.yml` have no model of their own, so each got a small frozen schema next to its loader: `SharedSection` (top-level `superusers`) and `ApiSection` (`context-path`, `enable-logging`).

Errors are the real win — a bad value now points at the file and line:

```
dature.errors.exceptions.FieldLoadError:   [port]  invalid literal for int() with base 10: 'not-a-number'
   ├── port: not-a-number
   │         ^^^^^^^^^^^^
   └── FILE '.../config/config.yml', line 28
```

## No new global state

- the module-level `dcf = Retort(...)` in the db parser is deleted
- sources, field mappings and type loaders are built per call, never at module level
- `dature.configure()` (which mutates library-wide config) is not called, and no `@dature.load` decorator is used — that form attaches a cached `Loader` to the class
- `dature.load()` in function mode builds a throwaway `Loader`; its retort cache is per-loader and `cache_engine` defaults off

One thing worth knowing: importing `dature` replaces `sys.excepthook` with its own (it renders `DatureError` without the traceback header and delegates everything else to the previous hook). That is unconditional in the library, with no opt-out — flagging it since it is a process-wide side effect the app now inherits.

## adaptix 3.0.0b9 → 3.0.0b12

dature requires `adaptix>=3.0.0b12`. Two behaviour changes in that range needed handling:

- `ABCProxy` now parametrises its target with the generic args of the abstract type, which only works when the abstract is still generic. `HintsList`/`Conditions` are already-parametrised `Sequence` subclasses, so they use a local `ConcreteProxy` that proxies to the type as given. (The existing "internal class, can be broken in next adaptix version" comments called this exactly.)
- sets dump as lists instead of tuples. `test_retort` and `test_serialize_simple` assert the new shape. Stored scenarios are unaffected — `yaml.safe_dump` and JSON render tuple and list identically.

## Behaviour

Unchanged. The source reads YAML 1.1 — the dialect PyYAML implemented — and env var expansion is off, so existing config files resolve exactly as before (a webhook secret like `my-$ecr3t` stays literal). Verified by parsing `config_dist/config.yml` and `tests/config/config.yml` with both the old and new loaders and diffing the resulting `ApiConfig`/`TgBotConfig` field by field: identical. Missing `mail`, `api.push` and `superusers` still fall back to the same defaults; `BotConfig.superusers` got `default_factory=list` so a config without the key keeps working as it did.

One regression to be aware of: each loader parses `config.yml` for itself, so a full `TgBotConfig` load costs ~83 ms against ~5 ms before (ruamel's pure-Python parser dominates). It is a once-per-process startup cost — both config providers are `Scope.APP` — but say the word if you would rather collapse the sections into fewer loads.

## Tests

`tests/unit/test_config_parser.py` covers the parsed bot and api configs: sections, the shared `superusers`, defaults for absent sections, the literal `$` in the webhook secret, and `token-expire-minutes` → `timedelta`.

`ruff format --check`, `ruff check`, `mypy` and `pytest tests/unit` (289 passed) are green locally. Integration tests need Docker, so CI has them.

---
_Generated by [Claude Code](https://claude.ai/code/session_016RsK5XdigrRBfY2Ltm6dWN)_


shvatka/api/app/config/parser/main.py
```diff
@@ -1,20 +1,33 @@
...
-        auth=load_auth(config_dct["api"]["auth"]),
-        context_path=config_dct["api"].get("context-path", ""),
-        enable_logging=bool(config_dct["api"].get("enable-logging", False)),
-        push=load_push(config_dct["api"].get("push"), dcf),
+        base=load_common_config(paths),
+        auth=load_auth(paths),
+        context_path=api.context_path,
+        enable_logging=api.enable_logging,
+        push=load_push(paths),
     )
```
bomzheg, line 29: can we do not enumerate all parts of config? can we load somehow whole config?

bomzheg: Yes — I tried it and a single `dature.load()` for the whole file works. But three things in the models can't come from `config.yml`, so what "whole config" costs depends on whether you want to move them.

**What blocks `dature.load(source, schema=ApiConfig)` today**

1. `Config.paths` — it's the input that *locates* `config.yml`, so it can't be a field loaded from it. I tried faking it (`type_loaders={Paths: lambda _: paths}` plus an alias onto some existing key) and it doesn't work: a `field_mapping` alias **moves** the source key rather than copying it, so borrowing `app` for `paths` just makes `app` missing. There is no in-memory/dict source in dature either — `EnvSource` reads `os.environ`, `ArgparseSource` wants a real `ArgumentParser` and reads `sys.argv`.
2. `Config.db` is typed as the `DBConfig` **Protocol**. adaptix can't build a loader for a Protocol, and it won't match a `loader()`/`type_loaders` predicate on one either — I checked against plain adaptix, without dature in the picture:
   ```
   ProviderNotFoundError: Cannot create loader for model.
     Location: ‹Holder.db: DBConfig›  → Cannot find loader
   ```
3. `TgBotConfig.tg_client` is derived from `bot.token`, not read from the file. `F[...].tg_client.bot_token: "bot.token"` does not work — a cross-level alias can't materialise a nested model that has no key of its own in the source.

The `api:` nesting, on the other hand, is **not** a problem — that one is solvable either way (`prefix="api"` plus `Absolute("app")`, `Absolute("db")`, … for the root sections, or a root-level load).

**Two ways forward, both verified against `tests/config/config.yml`**

*A — models untouched, one load per entry point.* One schema per entry point in the parser layer mirroring the file (everything except `paths`/`tg_client`, with `db` typed as `DBConfigProperties`), then assemble the real config from it. `load_config` drops from 9 parses to 1. The catch: that mirror dataclass duplicates the field list, so the enumeration doesn't disappear — it moves from the loader body into a schema. Honestly not obviously better than what's here now.

*B — the real fix, needs three model changes.* Drop `paths` from `Config` (DI already provides `Paths` on its own via `ConfigProvider.get_paths`; the `app_dir`/`config_path`/`log_path` properties would move to `Paths`), type `db` as `DBConfigProperties`, and build `TgClientConfig` in DI instead of in the config tree. Then it really is one line and no enumeration anywhere:

```python
def load_config(paths: Paths) -> TgBotConfig:
    return dature.load(
        config_source(paths, field_mapping={F[TgBotConfig].bot.superusers: "superusers"}),
        schema=TgBotConfig,
    )
```

B also kills the startup cost I flagged in the description — one parse instead of nine, so ~83 ms back down to ~10 ms.

I'd go with B, but it reaches into the config models, which is outside what you asked for in this PR — happy to do it here or as a follow-up. Tell me which and I'll push it. If you'd rather not touch the models at all, my read is that A isn't worth it and this should stay as is.

---

PR #342, opened 2026-08-10
Spawn every detached task on an app-scoped nursery

### LINKED ISSUE #156: Add nursery for async tasks (publish game and same
They should have self scope handling and acquire and finalize resources (like db session)

bomzheg:
Closes #156.

## The problem

`asyncio.create_task` was called straight from handlers (forum upload, scenario publishing to a channel, showing all hints of a level, the webhook background feed). Each of those tasks closed over request-scoped objects — a `HolderDao`, a `HintSender` — whose dishka scope is torn down the moment the handler returns, so the task kept working against a session that was already being finalized. Failures went into never-awaited tasks, and nothing at app level knew what was still running at shutdown.

## What this adds

A `Nursery` (`shvatka/core/interfaces/nursery.py`), provided once per app (`Scope.APP`) and taken as `FromDishka[Nursery]`:

```python
nursery.spawn(
    PublishScenarioToForumTask,
    PublishScenarioToForumParams(game_id=game_id, player_id=player.id, ...),
)
```

`AsyncioNursery` (`shvatka/infrastructure/nursery.py`) is the one place allowed to call `create_task`. Per spawned task it:

- opens a **fresh REQUEST scope**, so the task acquires and finalizes its own db session and every other request-scoped resource;
- holds a strong reference until the task ends;
- logs a failure instead of dropping it;
- cancels and awaits whatever is still running when the app closes.

A task is a pair, both in `shvatka/tgbot/tasks.py` and registered in `BackgroundTasksProvider`: a frozen params dataclass (the data of one run, handed to `spawn` and injected into the task's scope under its own type) and a DI-built class with an async no-arg `__call__`.

Params carry ids, not entities: the task reloads what it needs on behalf of the player who asked for it, through the new `PlayerIdentityProvider` (`shvatka/core/services/identity.py`) — an `IdentityProvider` for work with no telegram update or http request behind it. The spawning handler still runs the same authorization check eagerly, so a player without rights is told immediately instead of failing inside a task nobody is watching.

## Changes

- `core/interfaces/nursery.py` — `Nursery` / `BackgroundTask` protocols.
- `infrastructure/nursery.py` — `AsyncioNursery` + `NurseryProvider`, wired into `get_providers` so bot, api and root app all have it.
- `core/services/identity.py` — `PlayerIdentityProvider`.
- `tgbot/tasks.py` — `PublishScenarioToForumTask`, `PublishScenarioToChannelTask`, `SendLevelHintsTask`, `FeedUpdateTask` and their params, plus `BackgroundTasksProvider`.
- Call sites converted: `tgbot/handlers/game/organizer.py`, `tgbot/dialogs/game_manage/handlers.py`, `tgbot/dialogs/game_publish/handlers.py`, `tgbot/dialogs/level_manage/handlers.py`, `tgbot/utils/fastapi_webhook.py` (the background path also stopped passing an un-awaited `request.body` to `json_loads` and now answers `200` instead of `Response(content={})`).
- `GamePublisher.get_approximate_time_of(game)` — the estimate is told to the user before the publisher is built, which now happens inside the task.
- Ruff `TID251` bans `asyncio.create_task` / `ensure_future` / `TaskGroup` project-wide; the nursery carries the single `noqa`.
- The dishka container is now closed on app shutdown (root and api apps) — dishka's fastapi integration doesn't do it, and app-scoped finalizers (the nursery, the scheduler, `UserGetter`) hang off that close.
- `AGENTS.md` — the convention written down.

## Tests

- `tests/unit/test_nursery.py` — a spawned task runs in its own scope with its resource acquired and released, a failing task is contained and still finalized, `close()` cancels what is unfinished.
- `tests/unit/services/player_identity_test.py` — author resolves as primary org, secondary org is loaded once and cached, a stranger is not an org, and the provider has no team/chat.
- `tests/unit/test_di.py` already builds every container under `STRICT_VALIDATION`, which covers the new providers.

`ruff format --check`, `ruff check`, `mypy` and `pytest tests/unit` (300 passed) are green locally; integration tests need Docker and are left to CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_015fJ4Ko37W5DqFWTzjZ2gVP)_


AGENTS.md
```diff
@@ -165,6 +165,47 @@ So: an Interactor takes `identity: IdentityProvider` as a `__call__` arg (or a
...
+Why it matters: the nursery opens a **fresh REQUEST scope** per task, so the
+task acquires and finalizes its own db session (and views, clients, …) instead
+of borrowing the handler's, which is closed the moment the handler returns.
+The nursery also keeps a strong reference until the task ends, logs failures
+instead of dropping them, and cancels what is still running when the app shuts
+down.
+
+Two rules for writing one:
+
+- **Params carry ids, not entities.** The task loads what it needs itself, on
```
bomzheg: why? our entities are plain dataclasses, it's safe to pass them to any task or even thread

bomzheg: You're right — I had no good reason for it. DTOs are detached plain data, and reloading from ids bought nothing except a `PlayerIdentityProvider` to re-derive the authority the handler had already established.

Fixed in e3ecf76: params carry the loaded `game` / `level` again, `PlayerIdentityProvider` and its test are gone, and the rule now says what actually matters — entities travel in params, resources tied to the caller's scope (dao, session, `HintSender`) come from DI inside the task. `PublishScenarioToChannelTask` is back to receiving `game`/`game_stat`/`keys` from the handler and taking only `HintSender` from its own scope, which was the real bug: that sender's session died with the handler.

Net −176 lines.

---

shvatka/tgbot/utils/fastapi_webhook.py
```diff
@@ -118,13 +104,12 @@ async def handle(
         request: Request,
         bot: FromDishka[Bot],
         dispatcher: FromDishka[Dispatcher],
+        nursery: FromDishka[Nursery],
     ) -> Response:
         if not self.verify_secret(request.headers.get("X-Telegram-Bot-Api-Secret-Token", ""), bot):
             return Response(content="Unauthorized", status_code=401)
         if self.handle_in_background:
-            return await self._handle_request_background(
-                bot=bot, dispatcher=dispatcher, request=request
```
bomzheg, line 126: it was for multibots. lets revert here

bomzheg: Reverted in 4198139 — the file is now byte-identical to master.

You're right, and portability was the part I missed: routing its background updates through the nursery made this file `import shvatka.core.interfaces.nursery` and `shvatka.tgbot.tasks`, so it could no longer be pasted into another bot. The multibot angle would have bitten too — my task took `bot`/`dispatcher` as `FromDishka`, which resolves the container's singleton instead of the bot the request was resolved for, so the background path would have fed updates to the wrong bot.

To keep the `create_task` ban meaningful without touching this file, it gets a per-file `TID251` ignore with the reason written next to it, and AGENTS.md says to leave it self-contained and close to upstream. The `feed_update` task went with it — the handler was its only caller.

One thing I noticed while in there and deliberately did **not** touch, since it's pre-existing and this file is now hands-off: `_handle_request_background` passes `bot.session.json_loads(request.body)` without awaiting `request.body()`, and answers `Response(content={})`. That path only runs with `handle_in_background=True`, which this app never sets (`main_factory.py` passes `False`), so nothing hits it today. Happy to fix it in a separate PR if you want the portable copy to work out of the box for a bot that does enable it.

---

PR #351, opened 2026-08-12
Close dialogs whose state changed while the window waited

### LINKED ISSUE #339: check is user already not in team
```
async def on_start_my_team(start_data: dict, manager: DialogManager) -> None:
    dao: HolderDao = manager.middleware_data["dao"]
    player: dto.Player = manager.middleware_data["player"]
    team = await get_my_team(player=player, dao=dao.team_player)
    assert team
    manager.dialog_data["team_id"] = team.id
```

bomzheg:
Closes #339.

## The bug

A telegram window stays clickable forever, so the state a dialog was opened for can be gone by the time a button is pressed. Opening *🚩Моя команда* and pressing a button in it weeks later — after the captain removed the player — answered with an `AssertionError` on `assert team`, a dead button for the user and a traceback in the log chat.

The same shape was all over the team dialogs: `assert team` in `start_merge` / `gotten_chat_request` / `gotten_user_request`, `get_my_team_` returning `team=None` into a Jinja template that dereferences `team.name`, and a bare `done()` with no explanation in the rename / role / emoji handlers.

## The mechanism

`shvatka/tgbot/dialogs/outdated.py`:

- **`DialogOutdated`** — raised by a getter or a handler to say "what I was opened for is gone". Deliberately not an `SHError`: nothing went wrong in the domain, the window is simply old.
- **`OutdatedDialogMiddleware`** — an inner middleware of the dialogs router. Answers the callback query with an alert (or replies to the message, for `MessageInput`), then `manager.done()`.
- **`get_actual_team` / `get_actual_team_player` / `get_actual_teammate`** — guards that replace the hand-written checks, so each call site states which staleness it protects against.

Why a middleware and not an error handler: getters run inside `manager.show()` and cannot close a dialog themselves, and the catch-all handler on `dp.errors` wins propagation before any handler that would have a `dialog_manager` in its data. An inner middleware of a router wraps its sub-routers' handlers too, so one registration covers every dialog including getters. It has to be registered *after* `setup_dialogs` to sit inside aiogram_dialog's `ManagerMiddleware` — a unit test pins that ordering.

## Converted call sites

- `my_team_view` no longer captures a team id in `on_start`; `my_team_getter` resolves the current team on every render. Remembering an id and trusting it later is the bug this PR is about, so the dialog stopped doing it. The card body is shared with the public team view through `team_card()`.
- `captains_bridge` getters and handlers guard the acting player, their membership, and the selected player.

Also fixes a real mix-up found on the way: `change_permission_handler` took the selected player's *own* membership (`get_team_player_by_player`), so a player who had joined another team meanwhile would have had a permission flipped in that other team. It now resolves their membership in the captain's team.

## Tests

- `tests/unit/test_outdated_dialog_middleware.py` — notification and close for both callback and message events, no-context and no-manager cases, other exceptions pass through, and the middleware ordering invariant.
- `tests/integration/bot_full/test_outdated_dialog.py` — drives the real dispatcher with `BotClient`: opens the menu, removes the player from the team behind the dialog's back, clicks, and asserts the alert plus a re-rendered parent window without the stale button. Covers both the `on_start`-style path (my team) and the getter path (captain's bridge).

Docs: SHEP-0005 records the decision and the alternatives; `AGENTS.md` gets the rule that came out of it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GVMF9P95Jb6fq662kRgC4c)_

### DIFF
--- a/shvatka/tgbot/dialogs/outdated.py
+++ b/shvatka/tgbot/dialogs/outdated.py
@@ -1,0 +1,114 @@
+"""Guarding dialogs against state which changed while the window was waiting.
+
+Telegram keeps an inline keyboard clickable forever, so a window rendered today
+may get its button pressed in a month - by which time the captain has removed
+the player from the team the window was built for. Nothing tells the dialog
+that: `dialog_data` still holds the team id captured on start, and the getter
+still believes the player is a member. Re-checking on every event is the only
+option.
+
+`DialogOutdated`, raised from a getter or a handler, is how a dialog says "what
+I was opened for is gone". `OutdatedDialogMiddleware` catches it, tells the user
+what happened and closes the dialog, so they land back in a window built from
+current data instead of seeing an `assert` blow up.
+"""
+
+import logging
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from aiogram import BaseMiddleware
+from aiogram.types import CallbackQuery, Message, TelegramObject
+from aiogram_dialog import DialogManager
+from aiogram_dialog.manager.manager_middleware import MANAGER_KEY
+
+from shvatka.core.interfaces.dal.player import PlayerTeamChecker, TeamPlayerGetter
+from shvatka.core.models import dto
+from shvatka.core.players.player import get_full_team_player_or_none, get_my_team
+
+logger = logging.getLogger(__name__)
+
+NOT_IN_TEAM = "Ты больше не состоишь в команде. Окно устарело, открой меню заново"
+NO_MORE_TEAMMATE = "Этот игрок больше не состоит в твоей команде. Окно устарело"
+
+
+class DialogOutdated(Exception):
+    """The dialog was opened for state which no longer holds.
+
+    Not an `SHError`: nothing went wrong in the domain, the user simply pressed
+    a button of a window which reality has moved on from.
+    """
+
+    def __init__(self, notify_user: str, text: str = "") -> None:
+        super().__init__(text or notify_user)
+        self.notify_user = notify_user
+        self.text = text or notify_user
+
+
+async def get_actual_team(player: dto.Player, dao: PlayerTeamChecker) -> dto.Team:
+    """The team the player belongs to right now, whatever the window shows."""
+    team = await get_my_team(player=player, dao=dao)
+    if team is None:
+        raise DialogOutdated(NOT_IN_TEAM, f"player {player.id} is not in a team anymore")
+    return team
+
+
+async def get_actual_team_player(
+    player: dto.Player, team: dto.Team, dao: TeamPlayerGetter
+) -> dto.FullTeamPlayer:
+    """The acting player's own membership in `team`, as it is right now."""
+    team_player = await get_full_team_player_or_none(player=player, team=team, dao=dao)
+    if team_player is None:
+        raise DialogOutdated(NOT_IN_TEAM, f"player {player.id} is not in team {team.id} anymore")
+    return team_player
+
+
+async def get_actual_teammate(
+    teammate: dto.Player, team: dto.Team, dao: TeamPlayerGetter
+) -> dto.FullTeamPlayer:
+    """Membership of another player in `team`, as it is right now.
+
+    Guards the "selected player" of the captain's bridge: by the time the
+    captain presses a button the player may have left, or even joined another
+    team - and acting on their current membership would touch the wrong team.
+    """
+    team_player = await get_full_team_player_or_none(player=teammate, team=team, dao=dao)
+    if team_player is None:
+        raise DialogOutdated(
+            NO_MORE_TEAMMATE, f"player {teammate.id} is not in team {team.id} anymore"
+        )
+    return team_player
+
+
+class OutdatedDialogMiddleware(BaseMiddleware):
+    """Turns `DialogOutdated` into a notification and a closed dialog.
+
+    Installed as an inner middleware of the dialogs router, so it wraps every
+    dialog handler - including the window getters, which run inside `show()`
+    and therefore cannot close the dialog themselves.
+    """
+
+    async def __call__(
+        self,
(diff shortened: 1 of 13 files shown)


shvatka/tgbot/dialogs/outdated.py
```diff
@@ -0,0 +1,114 @@
...
+    async def _notify(self, event: TelegramObject, e: DialogOutdated) -> None:
+        if isinstance(event, CallbackQuery):
+            await event.answer(e.notify_user, show_alert=True)
+        elif isinstance(event, Message):
+            await event.answer(e.notify_user)
+
+    async def _close(self, manager: DialogManager | None) -> None:
+        if manager is None or not manager.has_context():
+            return
+        await manager.done()
```
bomzheg: may be reset stack and start menu?

shvatka/tgbot/dialogs/team_manage/getters.py
```diff
@@ -5,34 +5,35 @@
...
-        team = await get_my_team(player=player, dao=dao.team_player)
-        team_player = await get_full_team_player(player=player, team=team, dao=dao.team_player)
-    except PlayerNotInTeam:
-        team = None
-        team_player = None
-    return {"team": team, "team_player": team_player}
+    team = await get_actual_team(player, dao.team_player)
+    return {
+        "team": team,
+        "team_player": await get_actual_team_player(player, team, dao.team_player),
```
bomzheg: hmmmm? What's wrong with previous version? there was if, is dialog render ok for none?

shvatka/tgbot/dialogs/outdated.py
```diff
@@ -0,0 +1,114 @@
...
+    a button of a window which reality has moved on from.
+    """
+
+    def __init__(self, notify_user: str, text: str = "") -> None:
+        super().__init__(text or notify_user)
+        self.notify_user = notify_user
+        self.text = text or notify_user
+
+
+async def get_actual_team(player: dto.Player, dao: PlayerTeamChecker) -> dto.Team:
```
bomzheg: i am not sure this and next are useful functions

shvatka/tgbot/dialogs/team_manage/getters.py
```diff
@@ -52,10 +53,10 @@ async def get_selected_player(
     **_,
 ):
     player = await identity.get_required_player()
-    team = await get_my_team(player=player, dao=dao.team_player)
-    team_player = await get_full_team_player(player=player, team=team, dao=dao.team_player)
```
bomzheg, line 56: may be just get it from identity? and for playernotinteam catch and rethrow dialogoutdated?

bomzheg: Done, in db44784 — this and the "not sure these are useful" one below collapse into a single helper:

```python
async def get_actual_team_player(identity: IdentityProvider) -> dto.FullTeamPlayer:
    try:
        return await identity.get_required_full_team_player()
    except PlayerNotInTeam as e:
        raise DialogOutdated(NOT_IN_TEAM, f"player {e.player_id} is not in a team anymore") from e
```

`FullTeamPlayer` carries `.team`, so one call answers both "is there a team" and "am I still in it", and the two separate helpers are gone. It's also the same source the windows render from, and it caches within the request, so the extra query the old `get_my_team(player, dao)` added is gone too.

`get_actual_teammate` stays as a dao-taking function — identity only knows the acting player, and the point of that one is precisely to look up *someone else's* membership in *this* team.

---

shvatka/tgbot/dialogs/outdated.py
bomzheg: Agreed — the two are now one. See the reply on `team_manage/getters.py`: `get_actual_team_player(identity)` replaces both, since `FullTeamPlayer` already carries `.team`. Only `get_actual_teammate` is left beside it.

---

shvatka/tgbot/dialogs/outdated.py
bomzheg: Done in db44784 — `manager.start(states.MainMenuSG.main, mode=StartMode.RESET_STACK)`.

You're right that it's the better answer, and for a reason I'd underrated: `done()` only drops the dialog that noticed. Everything below it in the stack captured its own `dialog_data` against the *same* state that just turned out to be gone, so returning into the parent can just move the problem one window down. The main menu is the one window with nothing remembered — `get_main` rebuilds it from the identity provider on every render.

---

shvatka/tgbot/dialogs/team_manage/getters.py
bomzheg: Good push — I checked instead of assuming, and I had it wrong in the SHEP (fixed in db44784). Rendering `team=None` does **not** raise. jinja2's default `Undefined` swallows the attribute access:

```python
>>> await Jinja("🚩Команда: <b>{{team.name}}</b>").render_text({"team": None}, manager)
'🚩Команда: <b></b>'
```

So the old `try/except PlayerNotInTeam → team = None` produced a *silently* broken window rather than a traceback: `Капитанский мостик.` / `🚩Команда: <b></b>`, every permission-gated button gone (magic-filter resolves `F["team_player"].can_manage_players` on `None` to `None`, i.e. falsy), leaving only *🔀Перенести в другой чат* — which then hit `assert team` in `gotten_chat_request` — and *🔙Назад*. No hint of what happened.

That's the whole reason for the change: not that it crashed, but that it didn't. Now the user gets "Ты больше не состоишь в команде" and lands in a menu that matches reality.

One behaviour note while I'm here: the same window is where a captain who has temporarily left their own team would manage it (`check_can_manage_players` allows that explicitly). That case is not newly broken — `dao.get_team(player)` returned `None` for them before too, so they got the same empty window — but it is now a hard "you're not in a team" instead. Say the word if you'd rather I special-case the captain here.

---

PR #364, opened 2026-08-22
feat(game): hints of the passed levels

bomzheg:
## What

A playing team can now read back the hints it saw on the levels it has already left.

New endpoint: `GET /games/running/level/passed` →

```json
{
  "game_id": 7,
  "levels": [
    {
      "level_number": 0,
      "level_time_id": 11,
      "started_at": "...",
      "finished_at": "...",
      "hints": [ { "time": 0, "hint": [...] } ]
    }
  ]
}
```

It lists every level time of the team in the running game **except the current one**, each with the hints whose time had come between that level's start and the level up (`finished_at` is the start of the next level time).

## Why a separate endpoint

The client polls `/games/running/level/current` every few seconds, while the passed levels only change on a level up. Keeping them apart lets the UI fetch them once, on demand, when a player opens the spoiler.

## Hints the team never reached

Only hints that were actually published to the team are listed — a team that solved a level in three minutes doesn't get its 5-minute hint by passing it. The cdn file check follows the same rule: `GameFileReaderInteractor` now also accepts a guid that appears in the passed levels' hints, so the pictures keep loading after the level up, and nothing else becomes readable.

## Changes

- `core/games/dto.py` — `PassedLevelHints` / `PassedLevels`
- `core/games/adapters.py` — `GamePlayDao.get_passed_levels`
- `core/games/interactors.py` — `PassedLevelsReaderInteractor`; file access extended to the passed levels' guids
- `infrastructure/db/dao/rdb/level_times.py` — `get_team_level_times` (served by the existing `game_id, team_id, start_at` index)
- `infrastructure/db/dao/complex/game.py` — the implementation; resolving the current level time keeps the waiver check
- `infrastructure/di/interactors.py` — wiring
- `api/games/` — response + route

## Tests

- `tests/integration/test_game_play.py` — passed levels of a team on its second level, and the empty answer on the first one
- `tests/integration/api_full/test_game.py` — the endpoint's payload, and a photo from a passed level still served by the cdn endpoint

`ruff format --check`, `ruff check`, `mypy` and `pytest tests/unit` all pass locally; the integration suite needs docker, which isn't available in this environment, so it runs in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VkmvRZBrRnhQoMxta1fW5E)_

### DIFF
--- a/shvatka/infrastructure/db/dao/complex/game.py
+++ b/shvatka/infrastructure/db/dao/complex/game.py
@@ -4,7 +4,7 @@ import typing
 from dataclasses import dataclass
 from typing import Iterable
 
-from shvatka.core.games.dto import CurrentHintsOnly, Event
+from shvatka.core.games.dto import CurrentHintsOnly, Event, PassedLevelHints, PassedLevels
 from shvatka.core.interfaces.current_game import CurrentGameProvider
 
 from shvatka.core.interfaces.dal.complex import GamePackager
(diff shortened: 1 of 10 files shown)


shvatka/infrastructure/db/dao/complex/game.py
```diff
@@ -376,6 +376,30 @@ async def get_current_hints(self, identity: IdentityProvider) -> CurrentHintsOnl
             is_finished=is_finished,
         )
 
+    async def get_passed_levels(self, identity: IdentityProvider) -> PassedLevels:
```
bomzheg, line 379: we should have here waiver check. shouldn't we?

bomzheg: Yes, it's there — via the `get_level_time(identity)` call on the next line (line 383). `get_level_time` runs `check_waivers` and raises `WaiverError` before anything else happens, so a player without a waiver never reaches `get_team_level_times`. That's what the `# resolving the current level time also checks the waiver` comment is pointing at.

I've pushed `de17caf` adding `test_get_passed_levels_requires_waiver`, mirroring the existing `test_get_current_hints` waiver coverage, to pin this down explicitly for the new method.

---

Issue #185, opened 2026-02-16
check merge condition (may be will replace another dialogs)


bomzheg:
🚩 **process_level_result handler now appends raw conditions from timer dialog, potentially overwriting prior condition merging**

At lines 166-167 in the new code:
```python
if raw_conditions := result.get("conditions", None):
    manager.dialog_data["conditions"] = raw_conditions
```
This runs AFTER the merge logic at lines 140-165 which carefully merges keys, bonus keys, bonus hints, and routed conditions. If a result contains both `conditions` AND other keys like `keys` or `bonus_keys`, the merge logic would run first, then lines 166-167 would overwrite everything with the raw conditions from the timer dialog. This works correctly because the timer dialog only returns `{"conditions": ...}` without other key types, but the ordering makes the code fragile — a future dialog returning both would have the merge silently overwritten.

<a href="https://app.devin.ai/review/bomzheg/shvatka/pull/179" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open in Devin Review">
  </picture>
</a>

---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

_Originally posted by @devin-ai-integration[bot] in https://github.com/bomzheg/Shvatka/pull/179#discussion_r2813931115_

PR #366, opened 2026-08-22
perf: answer a key submission without waiting for telegram

bomzheg:
## Why

`POST /games/running/key` answered only after the whole game had been *told* about the key:

* `HintSender.send_hints` sleeps 1s after the caption and after every hint part — a level up with a three-part puzzle is 4s of sleeping alone;
* `WebPushSender.send_many` awaited one blocking `webpush` call at a time — a team of eight with two devices is 16 https round trips in a row, and again for the orgs;
* `BotOrgNotifier.notify` sends one message per org, sequentially; on the last team's last key `all_teams_finished` messages every team;
* all of it inside the transaction — `CheckKeyInteractor` committed *after* the fan-out.

The 1s sleeps are **not** touched — they are what keeps the bot inside telegram's per-chat limits.

## What changed

**An interactor decides what to show, as data, and shows it after the commit.**

```python
tasks = ShowTasks(view=key_tasks(new_key, input_container))
if level_up is not None:
    tasks.extend(level_up_tasks(level_up, ...))
await self.dao.commit()
await self.show(tasks)
```

A `ViewTask` is one thing the game has to show — `SendPuzzle`, `WrongKey`, `EffectsKey`, `GameFinished`, … — a frozen dataclass over domain dtos, collected in a plain list while the transaction is still open. Before the commit it is a list and nothing else, so **a transaction that does not land shows nothing**, by construction rather than by convention. `ShowTasks` holds one list per sender (view, org notifier, game log); order is kept inside a list and promised nowhere else.

**`GameView` collapses to one method**, `show(tasks)`. The eight `send_puzzle`/`wrong_key`/… methods stay as each implementation's private business, reached by a router that matches on the task type. `AnyViewTask` is a union, so a view that forgets a task fails `mypy` instead of silently ignoring it. Adding something to show is now one task class and one branch per view, instead of a method on the protocol times four implementations plus a forwarder in `ComplexView`.

**The api answers from the same list the game is shown from.** `insert_key` reads the tasks the interactor returned:

```python
tasks = await interactor(key=key.text, identity=identity, input_container=input_container)
shown_key = next((t for t in tasks if isinstance(t, KeyShown)), None)
```

`WebInput` is **deleted**. It existed only to smuggle a return value back out of a view — which is why the route had to `logger.critical` and answer 500 when a view had not been called. `InputContainer` now means only what its name says: what the key was typed into, so the bot knows which message to reply to. `WebGameView` is left with what it was really for (pushes) and is deferred along with the bot half.

**`ComplexView.show` spawns one nursery job for the whole list** (`shvatka/tasks.py`), so the messages of one request keep their order. The nursery hop is the view's business, not the interactor's — core says "show these" and does not know whether that happens now or later. The bot-only container binds `GameView` straight to `BotView` and renders inline, unchanged.

**Deferred does not mean dropped:**

* `AsyncioNursery.close` waits `drain_timeout` (15s) for running jobs before cancelling them, so an ordinary restart mid-fan-out no longer leaves a team with half a puzzle. Bounded, because publishing a scenario is minutes long and a restart cannot wait it out.
* Each task is delivered contained and retried when telegram says the failure was its own — `TelegramRetryAfter` (waiting exactly as long as it asked, capped), `TelegramNetworkError`, `TelegramServerError`; 3 attempts, doubling backoff. Being blocked by a chat or sending something malformed is not retried. Failures go to `BotAlert`, since no request is watching.

**Decide/commit/show on both paths.** `resolve_level_up` does the locking, reads and the one write and returns a `LevelUpOutcome`; `level_up_tasks` turns that outcome into tasks. Both are pure functions now, so what a key and a level up show is unit-testable without a database. `GamePlayTimerInteractor` follows the same order.

**Pushes go together** (`WebPushSender`) — up to `PARALLEL` (10) at once, expired subscriptions disabled after the gather on the request's session; only plain values (`_Recipient`) are handed to the sending thread.

### Accepted costs

* A retry restarts the whole task, so a failure halfway through a puzzle resends the parts that arrived. Deliberate — a duplicated puzzle is confusing, a half-sent one leaves the team with nothing to solve.
* An interactor that shows something must remember to call `show` after its commit. No finalizer catches what it forgot, on purpose: a flush that runs whatever happened would also run after a rollback.
* `OrgNotifier.notify` and `GameLogWriter.log` take a sequence now, so every caller wraps its event in a list. Mechanical, and it is what lets one request be one job per sender.
* A hard kill (SIGKILL, OOM) mid-delivery still loses the messages; durable delivery is SHEP-0010.

## Notes

* No migrations, no model changes, no config keys.
* The response payload is identical — the front-end needs no change.
* Net **345 lines less** than before this PR.

## Docs

* `shep-0009-key-submission-latency.adoc` — the decision, its two deviations, and what is still ahead.
* `shep-0010-message-outbox.adoc` — durable delivery, starting with an org resend button (shvatka-ui #185); its "shown but never committed" case is closed by this PR.
* `AGENTS.md` — the rules: decide as data, show after the commit, what a task may carry, and that a task may be rendered more than once.

## Tests

* `tests/unit/services/game_view_tasks_test.py` — `key_tasks` and `level_up_tasks` as pure functions: a duplicate is only told it is a duplicate, a wrong key, the three key types that count, a level up, one team finishing, and the last team finishing (congratulates every played team **and** writes the game log).
* `tests/unit/services/complex_views_test.py` — one job per request carrying the whole list; nothing to show starts no job; the caller is never kept waiting; orgs and game log likewise.
* `tests/unit/services/delivery_test.py` — retry policy: a dropped connection is retried and not alerted, giving up alerts once, being kicked from a chat is not retried, a failure is contained, broken alerting doesn't break delivery, flood-control wait honoured and capped, backoff grows.
* `tests/unit/services/nursery_close_test.py` — a short job finishes, a long one is cancelled once the drain is over and shutdown does not block on it.
* `tests/unit/services/web_push_sender_test.py` — pushes overlap, every subscription reached, 404/410 disables once, a provider error is not an expired subscription.
* `tests/integration/api_full/test_game.py` — a key is answered, and the message reaches telegram once the delivery job finishes.
* `tests/mocks/game_view.py` gained a router into the same recording lists, so the whole `test_game_play.py` suite exercises the new path unchanged.

`ruff format`, `ruff check`, `mypy` and `pytest tests/unit` are green locally (511 passed). **The integration suite needs docker and could not be run here — it is left to CI.**

## Before merging

Manual bot regression pass tracked in #376 — the sections this diff puts at risk are marked out there.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GGKUmpw4qhkHn1JnpyjQht)_


shvatka/core/games/interactors.py
```diff
@@ -364,6 +369,7 @@ async def __call__(
                 game=game,
                 at=now,
             )
+        # the level up may have finished the game for everyone
         await self.dao.commit()
```
bomzheg, line 378: i don't like two commits. may be view after commit? or why it's so?

bomzheg: Fixed in b4ffcca — one commit now, views after it.

Why it was two: the second commit only ever covered `dao.finish(game)`, the write that marks the game over when the last team finishes. It was unreachable before the first commit because `process_level_up` interleaved reads, writes and views — check `is_team_finished`, show it, check `is_all_team_finished`, write, show that. So a view always ran inside the transaction, and the write that happened after the first commit needed a second one to land.

Split into two halves instead:

- `resolve_level_up` — the global lock, the reads, and the one write, returning a `LevelUpOutcome` with everything a view will need afterwards (next level, level time to schedule against, spying orgs, the teams to congratulate).
- `show_level_up` — takes that outcome and only shows it; no dao access.

`CheckKeyInteractor` is now resolve → commit → show, so the key is confirmed first and the puzzle follows, same visible order as before. `GamePlayTimerInteractor` got the same treatment (its effects views moved after its commit too), so "no view runs inside a transaction" holds on both paths into a level up.

`AGENTS.md` and the SHEP (renumbered 0009, master took 0007/0008 meanwhile) carry the rule and the reasoning. Locally: ruff, mypy, 497 unit tests green.

---

shvatka/core/games/game_play.py
```diff
@@ -69,7 +71,7 @@ async def start_game(
         level_times[team.id] = await dao.set_to_level(team=team, game=game, level_number=0, at=now)
     await dao.commit()
 
-    await asyncio.gather(*[view.send_puzzle(team, game.levels[0]) for team in teams])
+    await view.show([SendPuzzle(team=team, level=game.levels[0]) for team in teams])
```
bomzheg: teams runs in parallel? not sequential (team1 hint1, team1 hint2, team2 hint1, team hint2)? we need parallel

shvatka/core/games/interactors.py
```diff
@@ -273,49 +330,58 @@ class GamePlayBaseInteractor:
     scheduler: Scheduler
     current_game: CurrentGameProvider
 
-    async def all_teams_finished(self, game: dto.FullGame) -> None:
```
bomzheg, line 276: make diff smaller in this file. i guess only some of lines should changes, and return value. most lines should be same

shvatka/core/games/game_play.py
bomzheg: You're right, and it was a real regression — master had `asyncio.gather` here and my batch was walked in one sequence, so twelve teams meant the last one waiting out the other eleven fan-outs at the start of a game.

Fixed in 719c5fe by stating the rule properly: order is a promise about **a chat**, not about the game. `group_by_team` (core/views/game.py) splits a batch into one list per team keeping each team's order, and both `BotView.show` and the combined delivery job `gather` the groups.

So within a team it stays sequential — a key is confirmed before the puzzle it opened — and across teams everything runs at once, as before. `KeyShown` grows a `team` property (`self.key.team`) so every task can be grouped the same way.

Three unit tests pin it, including that one team's tasks stay together when interleaved with another's.

---

shvatka/core/games/interactors.py
bomzheg: Agreed — most of that diff wasn't needed. Fixed in 719c5fe.

The cause was `LevelUpOutcome`: a dto whose only job was carrying data across the commit, which is exactly what the task list already does. Once it went, the restructure it forced went with it — `resolve_level_up`, `schedule_level_up` and the module-level mappers are gone, and `core/games/dto.py` is byte-identical to master again.

What's left in this file is what you described. `process_level_up`, `process_plain_level_up` and `all_teams_finished` keep their names, reads, writes and control flow; each returns `ShowTasks` instead of awaiting a view, so the only changed lines are the ones that used to show something:

```
-        await self.view.send_puzzle(team=team, level=next_level)
+        tasks.view.append(SendPuzzle(team=team, level=next_level))
```

Plus the return types, and one new `show` method that hands each list to its sender after the commit. `schedule_first_hint` is back inside `process_plain_level_up` where master had it, so the ordering matches master too.

`view_` stays where it was, now a staticmethod returning the tasks, so what a key shows is still unit-testable without a database.

The file is +80/−44 instead of a rewrite. The rejected `LevelUpOutcome` shape is written up in SHEP-0009 under Deviations so it doesn't come back.

---

shvatka/tgbot/tasks.py
```diff
@@ -25,13 +29,64 @@
...
+
+Delivery = Callable[[], Awaitable[None]]
+
+
+async def deliver(call: Delivery, alerter: BotAlert) -> None:
+    """Nothing watches a background send, so a failure is alerted, not raised."""
+    try:
+        await _with_retry(call)
+    except Exception as e:
+        logger.exception("cant deliver", exc_info=e)
```
bomzheg: write team id or chat id pls

bomzheg: Done in d67e114 — both ids.

Every deferred send now carries a description of what it was, and it goes into the alert, the exception log and the retry warning:

```
cant deliver SendPuzzle to team 7 (chat -1001234567890) because of Forbidden: bot was kicked
```

Built per sender: `f"{type(task).__name__} to team {team.id} (chat {team.get_chat_id()})"` for a view task, the number of orgs for a notice, the event type for a game log entry.

You were right that the old message was useless — a `BotAlert` saying only "Forbidden" tells whoever reads the log chat that something broke, but not which team is sitting without a puzzle, which is the one thing they need in order to go and resend it manually.

---

PR #371, opened 2026-08-23
feat(admin): let an admin change a game's status, and nothing else

bomzheg:
Backend half of bomzheg/shvatka-ui#164 — the admin panel could not walk a game
back, so waivers opened too early stayed open and the author could not edit the
game again.

## What it adds

- `GET /admin/games` — the games an admin may act on.
- `PUT /admin/games/{id}/status` — move one to another status.

## What an admin may see

Deliberately narrow, and the point of the change as much as the feature is:

- Only games that stopped being drafts appear — active (`getting_waivers`,
  `started`, `finished`) and `complete` (`ADMIN_MANAGEABLE_STATUSES`).
- Of them, only the status: both endpoints answer with `shared.Game` — id,
  author, name, status, start time, number — and the DAO behind them
  (`AdminGameStatusChangerImpl`) can reach nothing but the games table. No
  level, no hint, no file.
- A game in `underconstruction` or `ready` is reported as `GameNotFound`, the
  game an admin has just moved there included: the fix hands the game to its
  author and ends the admin's part in it.
- Nothing changes for the content rules that were already in place — a game
  that is not complete stays closed to admins, and the tests now pin that down
  for a *running* game specifically (the card, the author-facing card, the
  printed keys, the key log, the stat, the media).

## What follows a move

Leaving either out would undo the move:

- A game leaving the active statuses loses its planned start — `need_start_now`
  does not look at the status, so the scheduler would otherwise start the game
  minutes after the admin pulled it back.
- A game moving to `complete` goes through `complete_game`, the same domain
  service the author's own button uses (so it must be finished first, and gets
  its archive number). That service now keeps a number the game already has, so
  a round trip out of `complete` and back cannot renumber the archive.
- Moving *into* an active status checks that no other game is active — an admin
  repairing one game must not break the engine's one-active-game invariant.

## A rough edge the new tests turned up

Writing the running-game checks above surfaced a bug one step below them:
`get_typed_keys` and `get_game_stat` resolved the acting player's organizer with
`get_by_player`, which raises `NoResultFound` when there is no such row — so
somebody who is neither the author nor an org got an unhandled storage error (a
500) where the answer is a refusal. Nothing leaked; the request failed in the
wrong shape.

Looking that org up *is* the permission check, so `check_is_org` now turns "no
such org" into `NotAuthorizedForEdit` — the same refusal `check_can_read` and
`check_can_view_scenario` give for the same reason. `get_game_stat_with_hints`
(the spy screen) had it too and is fixed with it. Callers that already hold the
game are unaffected: a primary organizer is built from the game, never read from
storage, so the author's own screens and the bot's publish/manage dialogs take
the same path as before.

## Tests

- `tests/unit/services/admin_game_status_test.py` — 10 new unit tests over the
  interactors with in-memory fakes.
- `tests/integration/api_full/test_admin.py` — the list hides drafts and ready
  games, shows active and complete ones without content, the waivers → under
  construction round trip (with the planned start cancelled), the game
  disappearing afterwards, completing, numbering, the 403s, and the running-game
  content checks above.

Unit tests, `ruff`, `ruff format` and `mypy` pass locally. The integration tests
cannot run in this environment (testcontainers cannot pull its images through
the network policy), so CI is their first real run — the first one caught
exactly the rough edge described above.

https://claude.ai/code/session_01Wn1Yis79AC2cqyqGNz7Nda

### DIFF
--- a/shvatka/core/services/organizers.py
+++ b/shvatka/core/services/organizers.py
@@ -125,6 +125,23 @@ async def flip_deleted(manager: dto.Player, org: dto.SecondaryOrganizer, dao: Or
     await dao.commit()
 
 
+def check_is_org(org: dto.Organizer | None, player: dto.Player, game: dto.Game) -> dto.Organizer:
+    """The org, or a refusal for somebody who is not one.
+
+    A game that is not complete is readable by its author and its orgs, and by
+    nobody else — so looking the acting player's org up is itself a permission
+    check, and the answer to "no such org" is a refusal, not the storage error
+    of a row that isn't there.
+    """
+    if org is None:
+        raise exceptions.NotAuthorizedForEdit(
+            permission_name="game_org",
+            player=player,
+            game=game,
+        )
+    return org
+
+
 def check_can_see_log_keys(org: dto.Organizer):
     if not org.can_see_log_keys:
         raise PermissionsError(
--- a/shvatka/core/services/organizers.py
+++ b/shvatka/core/services/organizers.py
@@ -132,8 +132,14 @@ def check_is_org(org: dto.Organizer | None, player: dto.Player, game: dto.Game) 
     nobody else — so looking the acting player's org up is itself a permission
     check, and the answer to "no such org" is a refusal, not the storage error
     of a row that isn't there.
+
+    The org is checked to be *this* player's for *this* game rather than taken
+    on trust. Every caller looks it up by that pair, so a mismatch cannot come
+    from a user; it would mean the wrong row reached the check, and answering
+    permission questions from it is how one player reads another's game. The
+    refusal is the same either way — the caller is not an org of this game.
     """
-    if org is None:
+    if org is None or org.player.id != player.id or org.game.id != game.id:
         raise exceptions.NotAuthorizedForEdit(
             permission_name="game_org",
             player=player,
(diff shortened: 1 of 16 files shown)


shvatka/core/services/organizers.py
```diff
@@ -125,6 +125,23 @@ async def flip_deleted(manager: dto.Player, org: dto.SecondaryOrganizer, dao: Or
...
 
+def check_is_org(org: dto.Organizer | None, player: dto.Player, game: dto.Game) -> dto.Organizer:
+    """The org, or a refusal for somebody who is not one.
+
+    A game that is not complete is readable by its author and its orgs, and by
+    nobody else — so looking the acting player's org up is itself a permission
+    check, and the answer to "no such org" is a refusal, not the storage error
+    of a row that isn't there.
+    """
+    if org is None:
```
bomzheg: can we also add checks for org.player_id == player.id, and org.game_id == game.id?

bomzheg: Done in 117cd5b. `dto.Organizer` holds the objects rather than the ids, so it reads `org.player.id != player.id or org.game.id != game.id`, folded into the same condition as `org is None` — a mismatch gets the same refusal as not being an org at all.

Four unit tests in `tests/unit/services/check_is_org_test.py` cover it: an org of this game passes through, and `None`, another player's org, and an org of another game are each refused.

---

PR #374, opened 2026-08-24
feat(errors): ссылка на документацию в сообщении об ошибке

bomzheg:
An error the engine refuses an action with now names the documentation page that explains the rule, and each edge turns that into a link. The web ui side is [bomzheg/shvatka-ui#187](https://github.com/bomzheg/shvatka-ui/pull/187).

## What changed

**Core.** `SHError.doc_page` holds a `DocPage` (`shvatka/core/utils/doc_pages.py`) — a `StrEnum` whose value is the page path inside the docs' ROOT module, with an optional anchor (`player/play#keys`). It is a class attribute where every instance of an error means the same page (`InvalidKey` always means the key-entry rules), and a constructor argument for a one-off raise site — the same `PermissionsError` means the captaincy page in one place and the permissions page in another. The core never learns where the docs are published.

**Config.** A new `docs:` section carries `base-url`, `component` and `version`; `DocsUrlFactory` (`shvatka/common/docs.py`) builds `<base-url>/<component>/<version>/<page>.html` from it. Antora drops the version segment for the latest version, and `master` is published as that one, so `version` is **empty** by default (`/shvatka/setup_team/create_team.html`) and only a pinned release adds a segment (`/shvatka/3.7.0/...`). `config_dist` ships the pinned form.

**Bot.** `handle_sh_error` appends `ℹ️ Подробнее: …` to the error message. A callback query is answered with an alert, and an alert renders no html — so when such an error has a page, the link follows as a message of its own rather than being dropped.

**API.** `ErrorContent.doc_url` is serialized as `docUrl` next to the existing `type` / `text` / `description` — purely additive. New public endpoint **`GET /docs/pages`** hands the web ui every page keyed by the `DocPage` **member name**, with the url and title the engine built:

```json
{"pages": {"CREATE_TEAM": {"url": "https://…/setup_team/create_team.html",
                           "title": "Создание команды"}, …}}
```

The name is the contract, not the path: a page renamed in the docs changes the url here and the ui keeps linking to it. That is why the ui holds no page list and no docs configuration of its own.

**Docs.** SHEP-0007 records the decision (including both corrections made in review); `player/play.adoc` gains a `[#keys]` anchor so `InvalidKey` points at the key-entry section rather than the top of the page.

## Which errors carry a page

`InvalidKey` → ввод ключей, waiver errors → вейверы, team membership → вступление/выход, chat errors → создание чата, `PromoteError`/`CantBeAuthor` → аппрув, org errors → организаторы игры, scenario/level errors → как написать уровень, email errors → вход, plus a few raise sites in `core/services/team.py` (передача капитанства, перенос чата, полномочия). An error nobody mapped renders exactly as before, so the mapping can grow error by error.

## Testing

`pytest tests/unit` — 461 passed; `ruff format --check`, `ruff check`, `mypy .` clean. CI green, including the testcontainer `test` job.

New unit tests in `tests/unit/docs/`:
- every `DocPage` has a matching `.adoc` under `docs/modules/ROOT/pages/`, every anchor it names is declared in that file, every page has a nav title — a renamed or deleted page fails CI instead of shipping a dead link;
- the url factory: no segment by default, a pinned tag, a configured domain, a trailing slash, an anchor after the extension;
- the bot handler: the link in a message error, an extra message after a callback alert, silence for an error with no page;
- the API body: `docUrl` present, `null` when there is no page, the rest untouched;
- `/docs/pages` keyed by name rather than path.

Plus an integration test for the endpoint in `tests/integration/api_full/test_docs.py`, and `test_config_parser` asserting the docs section is read from the file.

The URL shape was verified by building the Antora site locally rather than assumed.


shvatka/common/config/models/main.py
```diff
@@ -24,6 +24,21 @@ class WebConfig:
...
+
+    The docs are an Antora site, so a page URL is
+    ``<base-url>/<component>/<version>/<page>.html``. ``version`` points at a
+    released tag rather than at ``master``: a link handed to a player has to keep
+    describing the engine they are using, not the one being written.
+    """
+
+    base_url: str = "https://bomzheg.github.io/Shvatka"
+    component: str = "shvatka"
+    version: str = "3.7.0"
```
bomzheg: bad default. let's use master as default. 3.7.0 is ok for config example

bomzheg: Done in 1f20fb1 — `version` defaults to `master`, `config_dist` keeps `3.7.0`. The docstring and SHEP-0007 now say why: the default should not age out as the engine moves on, and a deployment running a release pins its own tag.

---

PR #379, opened 2026-08-27
perf(dao): stop joining forum_users on every player load

bomzheg:
## Why

Every query that loaded a player also joined `forum_users` — 42 eager-load sites across 11 DAO modules, all funnelled through `models.Player.to_dto_user_prefetched()`.

Only six consumers ever read the forum identity:

| consumer | fed by |
|---|---|
| `api/players/responses.py` `PlayerWithIdentities` (`/me`, profile) | `get_by_id` / `get_by_user_id` / email login |
| `api/admin/responses.py` `AdminPlayer` | `search_players`, `get_by_id` |
| `core/players/player.py` `merge_players` guard | `get_by_id` |
| `core/search/interactors.py` `classify_player_hit` | `search_by_any_name` |
| `tgbot/dialogs/main_menu` link offer | current player |
| `Player.name_mention` for a forum-imported dummy | any |

Nothing in the bulk paths — game lists, level lists, waivers, key logs, organizers, team players, file infos — touches it. The shared `api/shared/responses.py` `Player` (game authors, team captains, poll entries) reads only `id`, `can_be_author`, `name_mention`, `username`.

The join was unconditional because `dto.Player` had no way to say "not loaded": `_forum_user is None` meant both "no forum account" and "we didn't ask", so loading it everywhere was the only safe option.

Two things made it redundant rather than merely unused. Migration `d4ddc1efb2ff` backfilled `players.username` for every row (`coalesce(tg username, forum name, 'id'||id)`), and every creation path since sets it via `get_free_and_associated_username` — so `name_mention` returns the username without consulting either identity, and a forum-only player is already found by theirs.

The cost was worst on the hottest path: `PlayerDao.get_by_id`, called for every authenticated API request from `api/app/dependencies/auth.py`, used two `selectinload` options — three roundtrips for a question most endpoints ask only for an id.

## What changed

**The forum identity gets its own dto.** `dto.Player` drops `forum_user`, `has_forum_user()` and `get_forum_name()`; `dto.PlayerWithForum(Player)` adds them back plus the `name_mention` branch for a forum-imported dummy. A subclass rather than a `NOT_LOADED` sentinel: a caller that needs the forum account says so in its signature and mypy rejects a plain `Player` there, instead of the check landing at runtime.

**Two ways to load a player by id.** `get_by_id` loads the telegram user only, with `joinedload` — one statement instead of three. `get_identities_by_id` loads both and returns `PlayerWithForum`, exposed as `PlayerIdentitiesGetter` (and `PlayerMergeOperandsGetter` for interactors needing each). `IdentityProvider.get_player()` still returns a plain `Player`, so `/me` pays for a second lookup instead of every request carrying the join.

**The bot asks for a flag, not the account.** The main menu offers "Былые свершения" only to players without a forum account; `ForumUserDAO.exists_for_player` answers that with one indexed lookup.

**Exports identify a player by username.** `export_stat.Player.from_dto` loses its `forum_name` fallback — the username branch already covers a forum-only player. `PlayerIdentity.forum_name` stays, so the crawler still produces it and old exports still import.

`search_players` keeps the join and returns `PlayerWithForum` for both callers, including the public search which discards it: the query is admin-shaped and already outer-joins `users` and `email_accounts`, so splitting it in two wasn't worth the duplication.

## Effect

Every list of games, levels, waivers, key logs, organizers, team players and files loses one `LEFT OUTER JOIN forum_users`; each authenticated API request loses two roundtrips. `/me` and the admin player card cost one extra query — the rare endpoint pays instead of the common one.

Verified by compiling the queries: the game, waiver and auth selects no longer reference `forum_users`; the profile select still does.

## Notes

- `docs/modules/shep/pages/shep-0011-player-identity-loading.adoc` records the decision, the rejected sentinel alternative, and a phase 2 for the same question about `_user`.
- A forum-imported dummy still carrying the `id{N}` placeholder username is mentioned as `dummy-{id}` on paths loading a plain `Player`; `renew_id_usernames` exists to remove those placeholders.

## Tests

- New `tests/unit/domain/test_player_forum_identity.py`, including a guard that the base `Player` does not grow the forum account back.
- Existing merge / search / `/me` mapper tests updated to the forum-aware dto.
- `ruff format`, `ruff check`, `mypy` and `pytest tests/unit` (531 passed) all clean locally. Integration tests need Docker, so they run in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UxaMRqNMVJUtGtisqQKXjY)_

### DIFF
--- a/shvatka/core/players/player.py
+++ b/shvatka/core/players/player.py
@@ -6,7 +6,6 @@ from itertools import pairwise
 from shvatka.core.interfaces.dal.player import (
     PlayerByIdGetter,
     PlayerByUserIdGetter,
-    PlayerIdentitiesGetter,
     PlayerPromoter,
     PlayerTeamChecker,
     PlayerUpserter,
@@ -72,12 +71,6 @@ async def get_my_team(player: dto.Player, dao: PlayerTeamChecker) -> dto.Team | 
 
 async def get_player_by_id(id_: int, dao: PlayerByIdGetter) -> dto.Player:
     return await dao.get_by_id(id_)
-
-
-async def get_player_identities_by_id(
-    id_: int, dao: PlayerIdentitiesGetter
-) -> dto.PlayerWithForum:
-    return await dao.get_identities_by_id(id_)
 
 
 async def get_player_by_user_id(user_id: int, dao: PlayerByUserIdGetter) -> dto.Player:
--- a/shvatka/core/players/player.py
+++ b/shvatka/core/players/player.py
@@ -6,6 +6,7 @@ from itertools import pairwise
 from shvatka.core.interfaces.dal.player import (
     PlayerByIdGetter,
     PlayerByUserIdGetter,
+    PlayerIdentitiesGetter,
     PlayerPromoter,
     PlayerTeamChecker,
     PlayerUpserter,
@@ -71,6 +72,12 @@ async def get_my_team(player: dto.Player, dao: PlayerTeamChecker) -> dto.Team | 
 
 async def get_player_by_id(id_: int, dao: PlayerByIdGetter) -> dto.Player:
     return await dao.get_by_id(id_)
+
+
+async def get_player_identities_by_id(
+    id_: int, dao: PlayerIdentitiesGetter
+) -> dto.PlayerWithForum:
+    return await dao.get_identities_by_id(id_)
 
 
 async def get_player_by_user_id(user_id: int, dao: PlayerByUserIdGetter) -> dto.Player:
@@ -373,8 +380,8 @@ async def dismiss_promotion(token: str, dao: InviteRemover):
 
 
 async def merge_players(
-    primary: dto.Player,
-    secondary: dto.Player,
+    primary: dto.PlayerWithForum,
+    secondary: dto.PlayerWithForum,
     game_log: GameLogWriter,
     dao: PlayerMerger,
     timeline: list[TimelineItem] | None = None,
--- a/shvatka/infrastructure/db/dao/rdb/waiver.py
+++ b/shvatka/infrastructure/db/dao/rdb/waiver.py
@@ -86,13 +86,9 @@ class WaiverDao(BaseDAO[models.Waiver]):
                 joinedload(models.Waiver.team).options(
                     joinedload(models.Team.chat),
                     joinedload(models.Team.forum_team),
-                    joinedload(models.Team.captain).options(
-                        joinedload(models.Player.user), joinedload(models.Player.forum_user)
-                    ),
+                    joinedload(models.Team.captain).options(joinedload(models.Player.user)),
                 ),
-                joinedload(models.Waiver.player).options(
-                    joinedload(models.Player.user), joinedload(models.Player.forum_user)
-                ),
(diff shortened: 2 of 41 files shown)


shvatka/core/players/player.py
```diff
@@ -73,6 +74,12 @@ async def get_player_by_id(id_: int, dao: PlayerByIdGetter) -> dto.Player:
     return await dao.get_by_id(id_)
 
 
+async def get_player_identities_by_id(
+    id_: int, dao: PlayerIdentitiesGetter
+) -> dto.PlayerWithForum:
+    return await dao.get_identities_by_id(id_)
```
bomzheg: do we have any reason to have this function?

shvatka/infrastructure/db/dao/rdb/waiver.py
```diff
@@ -153,9 +143,7 @@ async def get_played_teams(self, game: dto.Game) -> Iterable[dto.Team]:
                 joinedload(models.Waiver.team)
                 .joinedload(models.Team.captain)
                 .joinedload(models.Player.user),
-                joinedload(models.Waiver.team)
-                .joinedload(models.Team.captain)
-                .joinedload(models.Player.forum_user),
+                joinedload(models.Waiver.team).joinedload(models.Team.captain),
```
bomzheg: above we already loaded captain (and even it's user)

bomzheg:
Both review comments addressed in a48dbf6.

> [`core/players/player.py`](https://github.com/bomzheg/Shvatka/pull/379#discussion_r3880360567) — do we have any reason to have this function?

No — removed. `get_player_identities_by_id` was a one-line passthrough over the dao with a single caller. I added it only to mirror the `get_player_by_id` sitting right above it, but AGENTS.md asks for no new free service functions, so copying the old shape was the wrong instinct. `tgbot/handlers/merge.py` now calls `dao.player.get_identities_by_id(...)` directly, and the `PlayerIdentitiesGetter` import is gone from that module.

The `PlayerIdentitiesGetter` protocol itself stays — it is what the admin and notification interactors depend on.

> [`db/dao/rdb/waiver.py`](https://github.com/bomzheg/Shvatka/pull/379#discussion_r3887196274) — above we already loaded captain (and even it's user)

Correct, and it was a leftover of my own making. That block held two parallel chains, `team → captain → user` and `team → captain → forum_user`. Dropping the forum one truncated it to a bare `team → captain`, which ruff then collapsed onto one line — so it read like a deliberate option rather than a stump. Being a strict prefix of the chain beside it, it did nothing. Removed.

I scanned every options block under `dao/` for that same shape (a chain that is a prefix of another chain in the same block) and this was the only one.

Checks pass on the new head: `ruff format --check`, `ruff check`, `mypy`, `pytest tests/unit` (531).

Also opened #391 with a manual test checklist for the surfaces this PR can break silently — the merge guard in particular, since an unloaded forum identity there fails by *succeeding*.

Replies could not be posted on the threads themselves: the API rejects them with `user_id can only have one pending review per pull request`, so there is an unsubmitted review of yours sitting on this PR. Submitting or discarding it in the UI will free up in-thread replies; I have left both threads open rather than resolving them from outside.

---

PR #381, opened 2026-08-28
chore(lint): pay off most of the ruff/mypy ignore list

bomzheg:
## What

The `[tool.ruff.lint] ignore` list had 71 entries, split between "we
disagree with this rule" and a `# fixme` block that had stopped being
paid off. This turns off **42 of them** — 9 that already had no
violations, the rest fixed here — and rewrites what is left so each
remaining entry says *why* it is there.

Also drops two obsolete mypy overrides.

Lint, `ruff format --check`, mypy and `pytest tests/unit` are all green.

## One config fix worth reading

`[tool.ruff] target-version = "py312"` is new. ruff 0.2 cannot parse
`requires-python = ">=3.13"` and was silently linting the project as
**py38** — which is why `FA100`/`FA102` looked like they needed ignoring,
and why the py39+ rewrites never fired. Pinning it turns on `UP017`
(`datetime.UTC`), `UP038` (`X | Y` in `isinstance`), `UP006` and PEP 585
generics, all applied here. py312 is the newest version this ruff knows.

## Rules no longer ignored

**Already clean, just switched on:** `PD008`, `A003`, `PIE790`,
`RUF003`, `RET502`, `FA100`, `FA102`, `PLC1901`, `RUF006`.

**Fixed here:** `ANN002`, `ANN202`, `ANN204`, `ANN206`, `ARG005`,
`B905`, `BLE001`, `C408`, `C417`, `DTZ007`, `FBT003`, `I001`, `PGH003`,
`PGH004`, `PLR2004`, `PLW2901`, `RET503`, `RET504`, `RET505`, `RUF007`,
`RUF015`, `S106`, `S311`, `SIM102`, `SIM108`, `SIM114`, `SIM300`,
`TRY400`, `UP015`, `UP024`, `UP035`.

## Not purely mechanical — please look at these

- **`TRY400`** rewrites `logger.error` inside `except` blocks as
  `logger.exception`, so those lines now carry a traceback. Three call
  sites also stopped repeating the exception inside the message.
- **`ANN206` on the API response models.** `from_core` returns `None`
  for a `None` input, so a plain `-> Self` was a lie and `-> Self | None`
  broke every route that returns one. They now carry `@overload`s:
  `from_core(None) -> None`, everything else `-> Self`. This surfaced
  (and fixed) `Player.from_dto` in `export_stat`, which returns `Player`
  rather than `cls`.
- **`BLE001`.** Six `except Exception` blocks stay — jwt decoding, web
  push, pinning, username resolution — each now with a noqa saying why a
  blind catch is right at that boundary.
- **`SIM102`/`RET505`** touch control flow in the org/team filters and in
  `PlayerDao`'s username picking. Behaviour is unchanged, but they are
  the least mechanical diffs here.
- **`I001` is its own commit** (388 files, `--fix` plus `ruff format`,
  nothing else) so it can be reviewed or dropped separately.

## Deliberately still ignored

`UP040` is a **new** ignore: `type X = ...` builds a lazy
`TypeAliasType`, which is not callable — the autofix turned
`SHKey: TypeAlias = str` into one and broke `SHKey("SH123")` at runtime
(caught by the unit tests). `TCH001`/`TCH002`/`TCH003` stay off for the
same class of reason: dishka and pydantic read annotations at runtime.

`TRY301` moved from fixme to deliberate — the raise-in-try shape in the
auth flow is intentional.

Left in fixme, too big to be "easy": `ANN201` (612), `ARG001` (235),
`ANN401` (111), `PLR0913` (110), `ANN003` (63), `N818` (48, renames a
public exception API), `FBT001`/`FBT002` (~47 each), `ANN001` (32),
`ARG002` (22), `PLR0911` (8).

Tests and the forum crawler get per-file ignores rather than fixes, per
the request — a test hard-coding a fake secret or comparing against a
literal is not a smell, and the crawler is a one-shot import of another
engine's HTML.

## mypy

- `shvatka.infrastructure.di.*` no longer needs `call-overload` disabled.
- `matplotlib`, `aiogram_tests` and `diagrams` ship type information now,
  so they leave `ignore_missing_imports`. `telegraph`, `apscheduler`,
  `testcontainers`, `pywebpush` and `pillow_heif` still need it.

## Test plan

- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `mypy .` — 823 files, clean
- [x] `pytest tests/unit` — 523 passed
- [x] import smoke test over the DI, dialog, handler and model packages
      (import sorting can reorder side-effecting `__init__` imports)
- [ ] integration suite — needs docker, left to CI

https://claude.ai/code/session_01VMhWMCEDyRZYYF1M7JSAzu

---
_Generated by [Claude Code](https://claude.ai/code/session_01VMhWMCEDyRZYYF1M7JSAzu)_


shvatka/api/app/dependencies/auth.py
```diff
@@ -240,9 +240,8 @@ async def get_chat(self) -> dto.Chat | None:
             chat = await self.dao.chat.get_by_tg_id(cast(int, team.get_chat_id()))
             self.cache["chat"] = chat
             return chat
-        else:
-            self.cache["chat"] = None
-            return None
+        self.cache["chat"] = None
+        return None
```
bomzheg: i don't like this. previous code was two equivalent branches. now it have guard clause semantic. rollback same things and ignore this rule.

shvatka/core/services/key.py
```diff
@@ -194,18 +193,17 @@ async def process(
...
                 return [decision.effects]
-            elif isinstance(decision, action.MultipleEffectsDecision):
+            if isinstance(decision, action.MultipleEffectsDecision):
                 await self.find_and_process_level_up_if_needed(
                     team, lvl, decision.effects, now=now
                 )
                 logger.debug("found multiple effects %s", decision.effects)
                 return decision.effects
-            elif isinstance(decision, action.NotImplementedActionDecision):
+            if isinstance(decision, action.NotImplementedActionDecision):
```
bomzheg: rollbacke this and same. i've like elif for this cases. semanticly it is branches

shvatka/infrastructure/crawler/game_scn/loader/load_scns.py
```diff
@@ -113,12 +114,7 @@ async def set_results(game: dto.FullGame, results: GameStat, dao: HolderDao):
             player = await get_or_create_player(dao, key.player)
             await join_team_if_already_not(player, team, game_start_at, dao)
             await add_waiver_if_already_not(player, team, game, dao)
-            if i == len(keys) - 1:
-                is_correct = True
-            elif key.level != keys[i + 1].level:
-                is_correct = True
-            else:
-                is_correct = False
+            is_correct = bool(i == len(keys) - 1 or key.level != keys[i + 1].level)
```
bomzheg: as you touched these lines - let's make function from it. it don't understand what is is_correct and i =) 

add next comments:
to first if: last key always close the game
to elif: if level changed - key was correct
to else: otherwise not correct

shvatka/infrastructure/stored_scn_checker.py
```diff
@@ -31,8 +31,8 @@ async def main() -> None:
                 logger.info("checking game #%s %s (%s)", game.id, game.name, game.number)
                 try:
                     full = await dao.game.get_full(game.id)
-                except* exceptions.SHError as e:
-                    logger.error("got error %s %s", game.id, repr(e))
+                except* exceptions.SHError:
+                    logger.exception("got error %s", game.id)
```
bomzheg, line 35: explain

shvatka/api/app/dependencies/auth.py
bomzheg: Agreed, and rolled back everywhere — all 15 files RET505 had touched are byte-identical to master again, and RET505 is back in the ignore list, now in the deliberate section with the reason:

```toml
# an if/elif/else chain that returns from every arm is a set of parallel
# branches, not a guard clause — keep it readable as one
"RET505",
```

439fed5.

---

shvatka/core/services/key.py
bomzheg: Done — both decision chains here (`KeyProcessor` and `TimerProcessor`) and `define_next_level` are back to `elif`/`else`, along with every other RET505 site on the branch. The rule is ignored again, so it will not come back. 439fed5.

---

shvatka/infrastructure/crawler/game_scn/loader/load_scns.py
bomzheg: Extracted, with your three comments verbatim, in 439fed5:

```python
def is_key_correct(team_keys: Sequence[Key], index: int) -> bool:
    """Was the key at `index` the one that took the team off its level?

    The export only records the keys a team typed and the level it was on, so
    correctness has to be read back out of that sequence.
    """
    if index == len(team_keys) - 1:
        # last key always close the game
        return True
    elif team_keys[index].level != team_keys[index + 1].level:  # noqa: SIM103
        # if level changed - key was correct
        return True
    else:
        # otherwise not correct
        return False
```

Call site is now `type_=enums.KeyType.simple if is_key_correct(keys, i) else enums.KeyType.wrong`, so the `is_correct` local is gone and `i` only appears as the named `index` parameter.

The `# noqa: SIM103` is there because ruff wants the last two branches folded into `return team_keys[index].level != ...`, which would take the comments with it — the three cases spelled out is the point.

---

shvatka/infrastructure/stored_scn_checker.py
bomzheg: This is `TRY400`, then `TRY401`, in two steps.

Before:

```python
except* exceptions.SHError as e:
    logger.error("got error %s %s", game.id, repr(e))
```

`TRY400` says that inside an `except` block `logger.error` should be `logger.exception` — same ERROR level, but it attaches the exception being handled, so the log line carries the full traceback instead of nothing. That works inside `except*` too: within the handler the current exception is the matching sub-group, so you get the whole `ExceptionGroup` traceback, sub-exceptions included:

```
ERROR:t:got error 42
  + Exception Group Traceback (most recent call last):
  |   File "...", line 5, in <module>
  | ExceptionGroup: g (1 sub-exception)
  +-+---------------- 1 ----------------
    | ValueError: boom
    +------------------------------------
```

Once that traceback is in the record, `repr(e)` in the message is a strictly worse duplicate of its first line — which is what `TRY401` flags. Dropping it left `e` unused, hence `except* exceptions.SHError:` with no binding. So the log line lost `repr(e)` and gained the traceback.

Unrelated to this change, but I noticed it while reading the block: if `get_full` raises, `full` is never bound, so the `logger.debug("game %s, levels %s", full, full.levels)` on the next line raises `NameError` (or reuses the previous iteration's game) instead of moving on — the handler appends to `failed` but does not `continue`. Pre-existing on master and this is a one-shot script, so I left it alone; say the word and I will add the `continue`.

---

PR #389, opened 2026-08-30
perf(loop): measure the event loop, and take the blocking work off it

bomzheg:
During a game, endpoints that do almost nothing take seconds — `GET /notifications/unread-count`, one `COUNT` covered by the partial index `ix__notifications_unread`, has been observed at 5.4 s.

Nothing about those requests is slow. One process on one event loop serves the api and, through the webhook (`handle_in_background=False`), every telegram update — so whatever occupies that loop is added to the latency of everything else on it.

Four candidates produce exactly that symptom, and asgi-monitor's request duration cannot tell them apart: a blocked loop, an exhausted connection pool, a slow query, and uvicorn's own per-request overhead. So this lands the measurement first, and then the blocking work the measurement was going to find.

Design decision and rejected alternatives: `docs/modules/shep/pages/shep-0011-event-loop-latency.adoc`.

## Measurement

All on the global prometheus registry `setup_metrics` already exposes through `/metrics`.

- **`asyncio_loop_lag_seconds`** — a probe sleeps a fixed interval and records how much later than that it woke (`shvatka/common/loop_monitor.py`). When its p99 tracks the p99 of request duration, the loop *is* the queue.
- **`asyncio_loop_stalls_total`, and the traceback with it** — a watchdog thread watches a heartbeat the loop updates, and dumps `sys._current_frames()` for the loop thread when it goes stale, which *names the function* that blocked it. A thread rather than `loop.slow_callback_duration`, because that is only consulted in debug mode and debug mode costs more than the stalls it would find.
- **`db_pool_checked_out` / `_size` / `_overflow`, and `db_pool_checkout_seconds`** — from sqlalchemy pool events (`shvatka/infrastructure/db/metrics.py`). Gauges pinned at `pool_size + max_overflow` mean the pool is the other queue; the histogram says what put it there, since a connection held for seconds is one being carried across network io rather than across a query.

Configured under a new `monitoring:` section, every field defaulting to something safe on a live game, so the section may be left out.

Reading it: lag p99 tracking request p99 → the loop. Flat lag with the pool pinned → the pool. Neither → the query or the client.

## Blocking work moved to threads

There was exactly **one** `asyncio.to_thread` in the whole codebase (`api/app/utils/push.py`). These now run in threads:

| what | where |
| --- | --- |
| bcrypt hash/verify | `infrastructure/crypto/hasher.py` |
| matplotlib results picture | `infrastructure/picture/results_painter.py` |
| PDF keys sheet, xlsx export | `core/scenario/interactors.py` |
| file read/write, sha256, libmagic, HEIC transcode | `infrastructure/clients/file_storage.py` |

`PasswordHasher` became an **async protocol** as a result — the one source-compatibility break in this PR. Worth noting it is not only the login endpoint that paid for bcrypt: `ApiIdentityProvider.get_player` falls through to basic auth whenever a token is missing or invalid, so any request could.

For native code that releases the gil (bcrypt, `hashlib`, PIL, file io) a thread genuinely runs it beside the loop. For python code (matplotlib, openpyxl) the total cpu is unchanged — but one uninterruptible multi-second stall becomes slices of `sys.getswitchinterval()`, and the loop keeps being served. That is the larger win.

## Configuration and overhead

- **uvloop was never installed.** `pyproject.toml` depended on plain `uvicorn`, not `uvicorn[standard]`, and `lock.txt` — what the image installs — had no `uvloop` line, so production ran on the stock asyncio loop. Added as a direct dependency; `uvicorn.loops.auto` selects it purely on importability, so no code change. `lock.txt` was edited surgically (a full recompile would have bumped every transitive pin), and `uv.lock` is a two-line diff.
- **The engine was built with no pool arguments at all**, so sqlalchemy's defaults applied: 15 connections shared by the api, every webhook update and every nursery job — and a job opens a dishka scope of its own, so it holds one for as long as it runs. Now `pool-size` / `max-overflow` / `pool-timeout` / `pool-recycle` / `pool-pre-ping` under `db:`, **defaulting to what sqlalchemy already did** — this change alone moves nothing, it makes the ceiling visible and adjustable once the metrics say whether it is the one being hit. Sqlite is skipped; its pool takes no sizing arguments.
- **`show_game` fanned out to every team at once**, each a chain of telegram calls holding a session. Bounded to 8.
- **The request log ran through `BaseHTTPMiddleware`** (an anyio task group and a pair of memory streams per request) to produce a debug line. Rewritten as plain asgi.
- **`config_dist/logging.yml` had the root logger at `DEBUG`**, written synchronously on the loop thread. Now `INFO`.
- One eager `pprint.pformat` in `picture.py` that ran regardless of log level.

## A bug found on the way

**Starlette does not run the lifespan of a mounted sub-application.** `create_root_app` mounts what `create_app` builds, so a startup handler registered there never fires — verified against fastapi 0.111. The monitor is therefore registered on the **root** app by `setup_loop_monitor`, called from both entrypoints, and `tests/unit/test_loop_monitor.py` fails if it is moved back.

The same applies to `setup_application` in `create_root_app`, which registers aiogram's `emit_startup` / `emit_shutdown` on the mounted app — **those hooks do not run today**. Noted in the SHEP rather than changed here, since starting to run them is a behaviour change of its own.

## Deliberately not in this PR

Both will still show in the data; naming them so they aren't mistaken for a regression.

- **`GET /games/{id}/keys`** selects every key of a game, unpaginated, with seven nested `joinedload`s, and grows all night — which matches the spikes worsening toward morning. Largest single cpu consumer left, but fixing it means paginating or adding a cursor, which the org ui has to agree to.
- **The inline telegram webhook.** `handle_in_background=False` runs the whole aiogram pipeline inside the http request. Moving it changes what a delivery failure means.
- **Several uvicorn workers** — blocked on key submission being serialised by an in-process `asyncio.Lock` (`MemoryLockFactory`) and on `LevelTestingData` being in-process memory. A redis-backed `KeyCheckerFactory` is the prerequisite.

## Verification

`ruff format --check`, `ruff check` and `mypy` clean; **537 unit tests pass** (18 new: the monitor including a test that the watchdog traceback actually names the blocking function, the pool sizing and instrumentation, the new middleware, and the mounting regression). Integration tests need docker and will run in CI.

`tests/load/locustfile.py` now reproduces the shape of the problem: many users polling `/notifications/unread-count` — the canary, since nothing it does is slow — while a few hammer `POST /auth/token`. Before this change the poller's p99 tracks the login rate; after, it should not move.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XmocRWi8ugpFvoEPCyxKM8)_

### DIFF
--- a/shvatka/common/config/models/monitoring.py
+++ b/shvatka/common/config/models/monitoring.py
@@ -1,0 +1,30 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class MonitoringConfig:
+    """The ``monitoring`` section of config.yml — how closely the process
+    watches its own event loop, and how much of it it is allowed to hand to
+    threads.
+
+    Every field has a default that is safe on a live game, so the section may
+    be left out entirely.
+    """
+
+    probe_interval: float = 0.1
+    """how often the loop is asked for a turn, in seconds. also the resolution
+    of the watchdog: it cannot see a stall shorter than this."""
+
+    stall_threshold: float = 0.25
+    """a loop that hasn't had a turn for this long, in seconds, is stalled"""
+
+    stall_traceback: bool = True
+    """dump the stack the loop is stuck in when it stalls. costs one thread"""
+
+    stall_report_interval: float = 10.0
+    """at most one traceback per this many seconds. the counter still sees
+    every stall, so the rate stays honest while the log stays readable"""
+
+    blocking_threads: int | None = None
+    """size of the thread pool blocking work is handed to. ``None`` keeps
+    python's own default of ``min(32, cpu_count + 4)``"""
--- a/shvatka/common/config/models/monitoring.py
+++ b/shvatka/common/config/models/monitoring.py
@@ -3,28 +3,8 @@ from dataclasses import dataclass
 
 @dataclass
 class MonitoringConfig:
-    """The ``monitoring`` section of config.yml — how closely the process
-    watches its own event loop, and how much of it it is allowed to hand to
-    threads.
-
-    Every field has a default that is safe on a live game, so the section may
-    be left out entirely.
-    """
-
     probe_interval: float = 0.1
-    """how often the loop is asked for a turn, in seconds. also the resolution
-    of the watchdog: it cannot see a stall shorter than this."""
-
     stall_threshold: float = 0.25
-    """a loop that hasn't had a turn for this long, in seconds, is stalled"""
-
     stall_traceback: bool = True
-    """dump the stack the loop is stuck in when it stalls. costs one thread"""
-
     stall_report_interval: float = 10.0
-    """at most one traceback per this many seconds. the counter still sees
-    every stall, so the rate stays honest while the log stays readable"""
-
     blocking_threads: int | None = None
-    """size of the thread pool blocking work is handed to. ``None`` keeps
-    python's own default of ``min(32, cpu_count + 4)``"""
--- a/shvatka/common/config/models/monitoring.py
+++ b/shvatka/common/config/models/monitoring.py
@@ -3,6 +3,8 @@ from dataclasses import dataclass
 
 @dataclass
 class MonitoringConfig:
+    # off means nothing is started at all: no probe task, no watchdog thread
+    enabled: bool = True
     probe_interval: float = 0.1
     stall_threshold: float = 0.25
     stall_traceback: bool = True
--- a/shvatka/common/config/models/monitoring.py
+++ b/shvatka/common/config/models/monitoring.py
@@ -3,10 +3,9 @@ from dataclasses import dataclass
 
 @dataclass
 class MonitoringConfig:
-    # off means nothing is started at all: no probe task, no watchdog thread
-    enabled: bool = True
+    # on means a probe task, and with stall_traceback a watchdog thread too
+    enabled: bool = False
     probe_interval: float = 0.1
     stall_threshold: float = 0.25
     stall_traceback: bool = True
     stall_report_interval: float = 10.0
-    blocking_threads: int | None = None
--- a/shvatka/common/loop_monitor.py
+++ b/shvatka/common/loop_monitor.py
@@ -1,0 +1,145 @@
+"""Self-observation of the event loop: how long it goes without a turn, and why.
+
+The whole app — the api and, through the webhook, every telegram update — runs
+on one event loop in one process. So a request that does nothing but count rows
+can still take seconds: it waits its turn behind whatever else is occupying that
+loop. Request duration alone can't tell that apart from a slow query or a
+database connection that isn't free, which is what this module is for.
+"""
+
+import asyncio
+import logging
+import sys
+import threading
+import time
+import traceback
+from contextlib import suppress
+from types import FrameType
+
+from prometheus_client import Counter, Histogram
+
+from shvatka.common.config.models.monitoring import MonitoringConfig
+
+logger = logging.getLogger(__name__)
+
+LOOP_LAG = Histogram(
+    "asyncio_loop_lag_seconds",
+    "how much later than asked a periodic sleep on the event loop actually woke up",
+    buckets=(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0),
+)
+LOOP_STALLS = Counter(
+    "asyncio_loop_stalls_total",
+    "times the event loop went without a turn for longer than the stall threshold",
+)
+
+
+class LoopMonitor:
+    """Two views of the same thing: that the loop stalled, and what stalled it.
+
+    The lag histogram is the continuous one — a sleep that asks for a fixed
+    interval and records how much later than that it woke up. It costs one
+    timer per interval and is what a dashboard reads: when its p99 tracks the
+    p99 of request duration, the loop *is* the queue.
+
+    The watchdog is the one that names the culprit. It runs in a thread of its
+    own, watching a heartbeat the loop updates, and dumps the stack the loop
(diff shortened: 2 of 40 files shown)


shvatka/tgbot/tasks.py
```diff
@@ -38,6 +38,7 @@
 MAX_RETRY_DELAY: typing.Final = 30.0
 # what telegram may recover from on its own; everything else would fail the same
 RETRIABLE_ERRORS: typing.Final = (TelegramRetryAfter, TelegramNetworkError, TelegramServerError)
+PARALLEL_TEAMS: typing.Final = 8
```
bomzheg: at least 20:( if it's 9 team we can't show to 8 and after 30 seconds to last one

shvatka/common/loop_monitor.py
```diff
@@ -0,0 +1,114 @@
...
+    "how much later than asked a periodic sleep on the event loop actually woke up",
+    buckets=(0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0),
+)
+LOOP_STALLS = Counter(
+    "asyncio_loop_stalls_total",
+    "times the event loop went without a turn for longer than the stall threshold",
+)
+
+
+class LoopMonitor:
```
bomzheg, line 27: loop monitor is very complex, and scary me. let's add flag to disable it

shvatka/tgbot/tasks.py
bomzheg: You're right, and worse than the number: a slot is held for a team's whole chain, and `HintSender` sleeps a second between parts, so with a three-part puzzle the ninth team was waiting seconds even before a retry backed off up to thirty. Raised to 20, and the comment now says it is a ceiling against a runaway fan-out rather than a throttle.

While checking, the reason I introduced the bound turns out to be wrong. I wrote that each team's chain holds a db session — it doesn't: `show_game` takes one `view` from one dishka scope, so every team shares the single session that job opened. The fan-out never multiplied them. What is left as a real constraint is only Telegram's own rate limiting, which `_with_retry` already handles through `TelegramRetryAfter`.

So 20 is what you asked for, but the same objection returns at 21 teams, and I no longer have a good reason for any ceiling. Happy to drop the semaphore entirely and go back to gathering every team at once — say the word and it's a one-line revert.

---

shvatka/common/loop_monitor.py
bomzheg: Added `monitoring.enabled`, checked in `setup_loop_monitor` before anything is constructed — so off means no probe task and no watchdog thread, not idle ones. There's a test asserting neither appears.

```yaml
monitoring:
  enabled: false
```

There are two levels of off, in case only half of it worries you: `stall_traceback: false` drops just the watchdog thread (the complex half — the heartbeat, the stack dump, the rate limiting) and keeps `asyncio_loop_lag_seconds`, which is a ten-line sleep-and-measure loop. That gets you the metric that says whether the loop is the queue, without the thread.

Left the default at `true` since measuring is what the PR is for, but flipping it to opt-in is one line if you'd rather it stay off until you ask for it.

---

shvatka/common/config/models/monitoring.py
```diff
@@ -0,0 +1,12 @@
...
+
+@dataclass
+class MonitoringConfig:
+    # off means nothing is started at all: no probe task, no watchdog thread
+    enabled: bool = True
+    probe_interval: float = 0.1
+    stall_threshold: float = 0.25
+    stall_traceback: bool = True
+    stall_report_interval: float = 10.0
+    blocking_threads: int | None = None
```
bomzheg: this setting not related to monitoring. move it from this class to another.
also create pool will fine even when monitoring disabled.

shvatka/common/config/models/monitoring.py
```diff
@@ -0,0 +1,12 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class MonitoringConfig:
+    # off means nothing is started at all: no probe task, no watchdog thread
+    enabled: bool = True
```
bomzheg: default - false

shvatka/common/config/models/monitoring.py
bomzheg: Agreed, and the second half was a real bug — thank you. `set_blocking_threads` was called from inside the monitor's own startup, so turning the monitor off also stopped the pool being sized, and every offload in this branch lands in that pool.

Now `app.blocking-threads`, wired by its own `setup_blocking_pool` that both entrypoints call independently of the monitor:

```yaml
app:
  name: shvatka
  # blocking-threads: 16
```

Two tests in `tests/unit/test_blocking_pool.py` pin it — one asserts a thread from the pool is named `shvatka-blocking` with `monitoring.enabled: false`, the other that leaving the setting out keeps python's own sizing.

---

shvatka/common/config/models/monitoring.py
bomzheg: Done — `enabled: bool = False`. Nothing is constructed until a deployment asks for it.

The SHEP says so too now, and notes that reading a game's numbers means turning it on first.

---

shvatka/tgbot/tasks.py
bomzheg: Removed. `shvatka/tgbot/tasks.py` is byte-identical to master again, so the file leaves this branch's diff entirely.

You're right that it bought almost nothing: the db-session cost I introduced it for doesn't exist, so what was left was a semaphore that could make one team wait for another team's telegram calls — at the one moment where the whole point is that every team gets its level at the same time. Raising it to 20 moved the problem to the 21st team rather than answering it.

The SHEP's "Deviations from the plan" now records that the bound was written and taken out, and why, so it isn't reintroduced by someone reasoning from the same wrong premise.

---

PR #397, opened 2026-09-01
fix(game): plan levels and hints from the recorded level start

### LINKED ISSUE #383: время в самом начале интерактора брать и юзать его. чтобы не накапливались штуки типа 1 секунды в конце

bomzheg:
Closes #383.

## Problem

`event_wrapper` already takes `datetime.now()` on its first line, so moving the call earlier would not have helped much — the delay is not inside the interactor, it is *before* it. APScheduler wakes the job a fraction of a second after `run_date`, and that fraction used to become the start of the next level:

1. team is put on level N with `start_at = T`;
2. the level timer is scheduled for `T + action_time`, exactly;
3. the job actually runs at `T + action_time + δ` (δ ≈ tens or hundreds of ms);
4. `level_up(at=now)` writes `start_at(N+1) = T + action_time + δ`;
5. `schedule_first_hint(now=now)` plans level N+1's timer from that same late moment.

So δ went into the recorded start of every level and compounded. Over twelve timer-driven levels that is the reported ~1 s spread — teams that started together finished at 04:30:00 and 04:30:01, each carrying its own sum of scheduler jitter.

The same thing happened to hints inside a level: `send_hint` planned the next hint as "now + the gap between the two hints", where `now` was read *after* the DB round-trip and the view send, so every delivery pushed the rest of the level's hints a bit further out.

## Change

- `calculate_timer_level_up_time` (`shvatka/core/services/key.py`) — a level that ends by timer is recorded as ending when the timer was **due**: `started_at + action_time`, clamped so it can never be later than the wall clock. `LevelScenario.get_timer_action_time(effects_id)` looks the timer up by the id of the effects that fired, so a level with several timer conditions picks the right one; an effects id that belongs to no timer condition logs a warning and falls back to the old wall-clock behaviour.
- `process_plain_level_up` (`shvatka/core/games/interactors.py`) — the next level's first hint and its timers are scheduled from `lt.start_at`, the value just written, instead of from `now`. The `at`/`now` parameters of `process_level_up` / `process_plain_level_up` are gone, since the recorded start is the authoritative anchor. For a key-driven level up nothing changes: `start_at` *is* the moment the key was submitted.
- `send_hint` (`shvatka/core/games/game_play.py`) — the next hint is planned as `lt.start_at + hint.time`, via the new `calculate_hint_time`. Hint times are already defined as an offset from the start of the level, so this just stops re-deriving them from a drifting base. `calculate_next_hint_time` stays for level testing, which has no recorded start to anchor to.

Net effect: for a game where every level ends by timer, every team's level boundaries land on `game start + Σ action_time`, identical for all teams, no matter how the scheduler behaved on the night.

The scheduler's `misfire_grace_time` is 3600 s, so if a job does run very late the next `run_date` may be computed in the past and fires immediately — which is the catch-up behaviour we want, rather than pushing the whole game back.

## Tests

New `tests/unit/services/timer_drift_test.py`:

- the level-up time is the moment the timer was due, not the moment the scheduler woke;
- it is never recorded in the future (clamp);
- unknown effects fall back to the wall clock;
- twelve simulated timer levels with varying jitter land exactly on the sum of the timers;
- `get_timer_action_time` resolves by effects id and returns `None` for non-timer effects;
- `send_hint` plans the following hint at `level start + offset` even when it itself ran late.

`ruff check`, `ruff format --check`, `mypy` and `pytest tests/unit` (546 + 7 tests) all pass locally. The integration suite needs Docker, which isn't available here — it runs in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E7v2EbMvnJqj5UWwTgMkVB)_

### DIFF
--- a/shvatka/core/games/game_play.py
+++ b/shvatka/core/games/game_play.py
@@ -21,6 +21,9 @@ from shvatka.core.views.game import (
 
 logger = logging.getLogger(__name__)
 
+START_SNAP = timedelta(seconds=2)
+"""На столько планировщик имеет право промахнуться мимо назначенного старта игры."""
+
 
 async def prepare_game(
     game: dto.Game,
@@ -65,9 +68,12 @@ async def start_game(
     logger.info("game %s started", game.id)
     teams = await dao.get_played_teams(game)
 
+    started_at = snap_to_planned_start(game.start_at, now)
     level_times = {}
     for team in teams:
-        level_times[team.id] = await dao.set_to_level(team=team, game=game, level_number=0, at=now)
+        level_times[team.id] = await dao.set_to_level(
+            team=team, game=game, level_number=0, at=started_at
+        )
     await dao.commit()
 
     tasks = ShowTasks(view=[SendPuzzle(team=team, level=game.levels[0]) for team in teams])
@@ -187,6 +193,28 @@ def calculate_hint_time(level_started_at: datetime, hint: hints.TimeHint) -> dat
     задержки доставки будут копиться от подсказки к подсказке.
     """
     return level_started_at + timedelta(minutes=hint.time)
+
+
+def snap_to_planned_start(planned: datetime | None, now: datetime) -> datetime:
+    """
+    Начало игры — назначенное время, если планировщик проснулся почти вовремя.
+
+    Вся сетка игры отсчитывается от начала нулевого уровня, поэтому назначенное время
+    даёт ровные значения вместо «плюс сколько-то миллисекунд». Но проснуться можно
+    и сильно позже (например, после перезапуска), а тогда назначенное время означало бы,
+    что часть первого уровня уже прошла, — в этом случае берём фактическое.
+    """
+    if planned is None:
+        return now
+    if abs(now - planned) <= START_SNAP:
+        return planned
+    logger.info(
+        "game started at %s, %s away from planned %s, so planned start is not used",
+        now,
+        abs(now - planned),
+        planned,
+    )
+    return now
 
 
 def need_start_now(game: dto.Game) -> bool:
--- a/shvatka/core/games/game_play.py
+++ b/shvatka/core/games/game_play.py
@@ -74,7 +74,13 @@ async def start_game(
 
     await asyncio.gather(
         *[
-            schedule_first_hint(scheduler, team, game.levels[0], level_times[team.id].id, now)
+            schedule_first_hint(
+                scheduler=scheduler,
+                team=team,
+                next_level=game.levels[0],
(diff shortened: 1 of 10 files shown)


shvatka/core/games/game_play.py
bomzheg: also here. may be any other places?

bomzheg, line 172: how about timers? they time also from now

bomzheg: Timers here were already correct after the first commit — but only by accident of the argument's value, which is exactly why the comment was fair. `process_plain_level_up` passes `lt.start_at` into this parameter, so `now + condition.get_action_time()` was already "level start + action time". The parameter being called `now` hid that completely.

Fixed properly in 11093f7: the parameter is `level_started_at`, and `start_game` now passes `level_times[team.id].start_at` instead of its own wall clock, so the name and the value agree at every call site.

---

bomzheg: Done, and I went through the rest. In 11093f7:

- `calculate_first_hint_time` now goes through `calculate_hint_time(level_started_at, hint(1))` instead of `now + (hint1.time - hint0.time)`. Same number today (hint 0 is always at minute 0 — `HintsList.verify` requires it), but it's the offset-from-level-start form, not a delta from a moving base.
- `calculate_next_hint_time` and `calculate_next_hint_timedelta` are gone. They had no callers left, and they are the exact shape that produced the drift, so leaving them around is an invitation to reintroduce it. One way to compute a hint time now: `calculate_hint_time`.
- **Level testing had the same bug.** `send_testing_level_hint` planned the next hint from the wall clock at the moment the previous one was delivered, so an org testing a level saw the same creep. It now anchors to the recorded start of the test, via a new `get_started_at` on `LevelTestProtocolDao` (the start was already stored in the in-memory bucket, it just had no getter).

Two places I looked at and deliberately left alone:

- **`start_game`.** It takes `datetime.now()` and uses it both as the start of level 0 and as the anchor for its schedule, so it's internally consistent and all teams share one value — it can't produce a spread between teams. Anchoring to `game.start_at` instead would make the timeline land on exact planned times, but `need_start_now` accepts a start up to 30 minutes late (server restart, etc.), and then level 0's timers would already be expired the moment the game began. Tell me if you'd rather have the exact anchor with a small tolerance and I'll add it.
- **`CheckKeyInteractor`.** `now` there is when the key was actually submitted. That one really is wall-clock time and should stay.

Reading path (`GamePlayReaderInteractor`, `PassedLevelsReaderInteractor`) compares `now` against a stored `start_at` to decide what to show — nothing accumulates there.

---

PR #401, opened 2026-09-02
Check Telegram takes a file: strict on zip import, force on web upload

### LINKED ISSUE #357: при сохранении сценария через API может сразу чекнуть что все файлы отправляются в тг?
и вернуть список ошибок. 
дать юзеру выбор - либо сохранять как есть и пофиг на всё ну либо исправляй

bomzheg:
## Summary

A hint reaches a team as a Telegram message, so a file Telegram refuses is a hint that can never be shown. Both ways a file arrives now find that out while the author is still looking, instead of during a game — but they answer a refusal differently, because they are different situations:

- **Zip import stays strict.** A package is expected to be correct, so `upsert_files` tries every file and refuses the whole import if Telegram would not take any of them, raising `FilesCantBeSentToTg` that names all of them at once (the author fixes the package in one pass instead of one file per attempt). No force — a half-deliverable package is not worth saving. The bot's zip dialog prints that list.
- **A web upload takes one file at a time**, and normally it just works. `UploadGameFileInteractor` now sends the file to Telegram before committing; a refusal fails the upload and keeps nothing, unless the author passes `force=true` — the rare deliberate case — which stores it without a `file_id`, to be sent by content.

`FileGateway.put` therefore keeps its old signature (it raises), and `FileRejectedByTelegram` / `FilesCantBeSentToTg` compose their own user-facing message from the filename and what Telegram said, so every edge that shows `notify_user` says which file to fix.

Also adds the zip endpoints the web needs to move a whole game (used by bomzheg/shvatka-ui#196):

- `POST /games/my/zip` — import a package as a new draft, or rewrite your own game of that name
- `GET /games/my/{id}/scenario/zip` — export the scenario with its media

## Context

Closes #357. The earlier version of this branch had force on the zip import and nothing on the web upload; this is that inverted, per review.

## Test plan

- [x] `ruff format --check .`, `ruff check .`, `mypy .` — clean
- [x] `pytest tests/unit` — 580 passed, including new unit tests for the aggregation, the upload interactor's force behavior, and the gateway's translation of an aiogram error
- [x] Integration tests added: the cdn upload (accepted / refused / forced / nothing half-written) and the zip endpoints (export, round-trip import, refused package, not-a-zip). They run against a new fake Telegram (`FileGatewayMock` + `FakeTelegram`) so a test can have a file accepted or refused; no Docker in this sandbox, so CI is their first real run
- [ ] Bot regression pass for the zip-import dialog (`underconstruction` states)

https://claude.ai/code/session_01Uu1EoNPkZHJUD2Aj2opr49

### DIFF
--- a/shvatka/core/services/game.py
+++ b/shvatka/core/services/game.py
@@ -53,15 +53,13 @@ async def upsert_game(
     dao: GameUpserter,
     retort: Retort,
     file_gateway: FileGateway,
-    force: bool = False,
 ) -> dto.FullGame:
-    """Save the scenario, failing up front if telegram would reject one of its files.
+    """Save an uploaded scenario package, files and all.
 
-    ``force=False`` (the default) checks every file can reach telegram before
-    anything is saved, raising ``FilesCantBeSentToTg`` with the full list of
-    problem files if not. ``force=True`` saves regardless — a file telegram
-    rejected is kept without a ``file_id`` and sent by content instead, the
-    first time it is shown in a game.
+    Every file is sent to telegram before anything is saved; if telegram
+    refuses any of them, nothing is written and ``FilesCantBeSentToTg`` names
+    all the problem files at once. A package is imported whole or not at all —
+    there is no saving it half-deliverable.
     """
     check_allow_be_author(author)
     game_scn = parse_uploaded_game(raw_scn, retort)
@@ -72,7 +70,7 @@ async def upsert_game(
             )
         game = await dao.get_game_by_name(name=game_scn.name, author=author)
         check_game_editable(game)
-    guids = await upsert_files(author, raw_scn.files, game_scn.files, dao, file_gateway, force)
+    guids = await upsert_files(author, raw_scn.files, game_scn.files, dao, file_gateway)
     check_all_files_saved(game=game_scn, guids=guids)
     game = await dao.upsert_game(author, game_scn)
     await dao.unlink_all(game)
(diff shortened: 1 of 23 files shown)


shvatka/core/services/game.py
```diff
@@ -54,6 +54,13 @@ async def upsert_game(
     retort: Retort,
     file_gateway: FileGateway,
 ) -> dto.FullGame:
+    """Save an uploaded scenario package, files and all.
+
+    Every file is sent to telegram before anything is saved; if telegram
+    refuses any of them, nothing is written and ``FilesCantBeSentToTg`` names
+    all the problem files at once. A package is imported whole or not at all —
+    there is no saving it half-deliverable.
+    """
```
bomzheg: to this file was only docsting added?! don't make changes not related to pr

bomzheg: Right, that was the only change here — a docstring. Dropped it, `game.py` is now untouched by this PR.

For the record on why it was there: `upsert_game` does change behaviour in this PR, but through `upsert_files` — that call now refuses the whole save when telegram won't take a file. That is documented on `upsert_files` itself, in `core/services/scenario/files.py`, which this PR does change.

---

shvatka/tgbot/dialogs/game_manage/handlers.py
bomzheg: let's inline this function

bomzheg: Done — `common_show_zip` is gone, both handlers answer the document themselves. Three lines each now that the interactor holds what the helper used to.

---

PR #411, opened 2026-09-05
feat(tgbot): send prepare messages before unpinning the previous game

### LINKED ISSUE #386: сначала отправить команде препаре гейм а потом отпинивать прошлую игру
на случай рейтлимита - чтобы он попал на unpin.

also we should have to check is sleep exists for unpins

bomzheg:
Closes #386

## What

Preparing a game does two things for every team: it tells the team the game is coming, and it cleans up the pins a previous, abnormally finished game could have left. The cleanup went first, so a flood limit hit while unpinning could cost a team its prepare message — the one part of the preparation nobody can do without.

- `BotView.prepare_game_view` now sends every team's prepare message (and drops the waivers keyboard) first, and unpins the previous game afterwards, in a second pass over the same chats. A flood limit hit lands on the unpins, which are best-effort.
- The per-team part moved into `BotView._prepare_team`, and the `game.start_at` check moved out of the loop — it doesn't depend on the team and used to abort the whole preparation from inside the first iteration anyway.

## Is there a sleep for unpins? (the second half of the issue)

There was one, but only *within* a single `unpin` call: `unpin_all` fires the level and the bonus category one after another, and preparing a game now unpins every team in a row — telegram counts all of those the same. `MessagePinner` now spreads unpins out **per chat** (`unpinned_chats`) rather than per call, so `SLEEP` also applies between the two categories.

## Tests

- `tests/integration/bot_full/test_game_prepare.py` — the prepare message reaches the team before the previous game is unpinned.
- `tests/integration/bot_full/test_pinner.py` — unpins of both categories in a row stay spread out.

`ruff check`, `ruff format --check` (pinned 0.2.x) and `mypy` are clean. The suites weren't run locally: the environment can only build the venv on 3.14.0rc2, where pydantic fails to import — CI runs them.

https://claude.ai/code/session_01UbSkFo95cWtAefdm7gGj88

---
_Generated by [Claude Code](https://claude.ai/code/session_01UbSkFo95cWtAefdm7gGj88)_

### DIFF
--- a/shvatka/tgbot/views/pinner.py
+++ b/shvatka/tgbot/views/pinner.py
@@ -3,7 +3,7 @@ import enum
 import logging
 import typing
 from collections.abc import Iterable
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from datetime import timedelta
 
 from aiogram import Bot
@@ -28,17 +28,12 @@ class MessagePinner:
     dao: PinnedMessageDao
     rights: BotRights
 
-    unpinned_chats: set[int] = field(default_factory=set, init=False)
-    """Chats this pinner already unpinned in - the next unpin there waits :attr:`SLEEP`.
-
-    Unpins are spread out per chat, not per call: preparing a game unpins both
-    categories of every team in a row, and telegram counts them all the same.
-    """
-
     SLEEP: typing.ClassVar[timedelta] = timedelta(seconds=1)
-    """Between unpins: a level's worth of them at once is flood control.
+    """Before every unpin: a level's worth of them at once is flood control.
 
-    Pins need none — they follow sends that are already a second apart
+    Preparing a game unpins both categories of every team in a row, so waiting
+    only between the unpins of one call would not spread them out at all.
+    Pins need no wait — they follow sends that are already a second apart
     (:class:`~shvatka.tgbot.views.hint_sender.HintSender`), while a level up
     unpins everything the level pinned in one go.
     """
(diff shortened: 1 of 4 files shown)


shvatka/tgbot/views/pinner.py
```diff
@@ -76,9 +83,10 @@ async def unpin(self, chat_id: int, category: PinCategory) -> None:
                 "can't get pinned messages (%s) of chat %s", category.value, chat_id, exc_info=e
             )
             return
-        for number, message_id in enumerate(message_ids):
-            if number:
+        for message_id in message_ids:
+            if chat_id in self.unpinned_chats:
```
bomzheg: omg. just sleep every time. we do not need unpinned_chats at all. and this if also.

bomzheg: Done in a67b605 — `asyncio.sleep(SLEEP)` before every unpin, `unpinned_chats` and the `if` are gone. The `SLEEP` docstring now says "before every unpin", and `test_unpins_are_spread_out` expects three waits for three messages; the cross-category test became redundant and was dropped.

---

shvatka/tgbot/views/pinner.py
```diff
@@ -29,9 +29,11 @@ class MessagePinner:
...
-    """Between unpins: a level's worth of them at once is flood control.
+    """Before every unpin: a level's worth of them at once is flood control.
 
-    Pins need none — they follow sends that are already a second apart
+    Preparing a game unpins both categories of every team in a row, so waiting
+    only between the unpins of one call would not spread them out at all.
+    Pins need no wait — they follow sends that are already a second apart
     (:class:`~shvatka.tgbot.views.hint_sender.HintSender`), while a level up
     unpins everything the level pinned in one go.
     """
```
bomzheg: remove docstring.

bomzheg: Removed in a666ea8 — `SLEEP` is left bare.

---

PR #414, opened 2026-09-07
docs: SHEP-0003 season schedule — the plan, revised

bomzheg:
Rewrites `docs/modules/shep/pages/shep-0003-season-schedule.adoc` and **implements nothing**. The engine, bot and web-UI code this branch carried earlier is reverted (the revert keeps it in history, so each phase stays cherry-pickable); bomzheg/shvatka-ui#201 is reverted the same way. #167 gets built next, one phase per pull request.

## Why

The first attempt implemented the whole SHEP in one go, and reviewing it turned up decisions the plan had wrong rather than code the plan had right. Those decisions belong in the document before anything is written against them again.

## What changed in the plan

**Permissions.** Any author (`can_be_author`) may compose, publish, edit and take — a date someone else took included. No owner lock and no superuser override, so no `NotSlotOwner` and no `SlotAlreadyTaken`. The rationale is in the rewritten `Slot ownership` section: promotion is the whole gate, owning a date is intent rather than access, and a superuser acting in the ordinary schedule is acting as an author like anyone else. Every change is recorded and reversible, which is what makes a shared date safe.

**Admin panel** — a new section and a new phase 4. Putting *any* player or *any* team on a date is an admin-panel action, behind `PUT /admin/seasons/{year}/slots/{id}/author` and an `/admin/season` page. The ordinary schedule view is byte-for-byte the same for an admin and for everyone else.

**Web UI.** A tap opens a modal and never writes — for a free date and a taken one alike. Under 480px the modal is a bottom sheet, replacing the panel below the calendar, which a phone could not use.

**Bot.** Tapping never writes there either; a date's window is where anything happens. The team picker offers captained teams to admins too.

**Phasing.** Four phases now, and *one phase, one pull request*.

**Traps.** The four things CI caught the first time are written down so the next attempt does not re-find them: a dataclass field named after its own type breaking pydantic's annotation evaluation; loader options that stop one level short of what `to_dto` walks; reading a change payload after the write that changes it; and the `aiogram_dialog` import that makes mypy mis-resolve a type in an untouched file.

**Open questions** — both settled: one digest line per date, and discard-and-reload on a compose-mode conflict.

## Testing

Documentation only — no code changes to test.

https://claude.ai/code/session_01VFYy6TmPfofatvKwp4jNfP


shvatka/core/games/editor_interactors.py
```diff
@@ -164,22 +165,29 @@ class PlanGameStartInteractor:
     dao: GameStartPlanner
     scheduler: Scheduler
     game_log: GameLogWriter
+    slot_sync: SyncLinkedSlotInteractor
```
bomzheg: it's bad if interactor depends on another interactor.
things like this should be as core functions

shvatka/core/season/adapters.py
```diff
@@ -0,0 +1,45 @@
...
+    SlotOrgWriter,
+    ScheduleChangeReader,
+    ScheduleChangeWriter,
+    SeasonAudienceReader,
+    NotificationWriter,
+    Protocol,
+):
+    """Everything a season use case may need, behind one Protocol.
+
+    `AGENTS.md` allows an interactor at most one dao, so the four season tables
```
bomzheg: remove all new comments and docstrings

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,628 @@
...
+        if season is None:
+            raise exceptions.SeasonNotFound(text=f"no season for {year}")
+        return season
+
+    @staticmethod
+    def _get_slot(season: season_dto.Season, slot_id: int) -> season_dto.Slot:
+        for slot in season.slots:
+            if slot.id == slot_id:
+                return slot
+        raise exceptions.SlotNotFound(text=f"season {season.year} has no slot {slot_id}")
```
bomzheg: this should be method of Season

PR #417, opened 2026-09-15
feat: season schedule in the engine (SHEP-0003 phase 1)

bomzheg:
Phase 1 of `docs/modules/shep/pages/shep-0003-season-schedule.adoc` — the engine half of #167. Phases 2 (bot), 3 (web) and 4 (admin panel) each get their own pull request; this one is their only prerequisite, and they may land in any order after it.

## What it does

A **season** is one calendar year's plan of games as a list of **dates**. There is no draft: the client asks for the nine default dates (the first Saturday strictly after 9 May, then every 21 days), edits them locally, and publishes the whole list in one request. Publication writes the season and its dates in one transaction, announces it in the game-log channel as a pinned message, and writes a `low`-severity `season_schedule_changed` notification to everyone who was in a team — or organizing a game — during the current or previous year.

Every later edit records a `season_changes` row **in the same transaction as the write it describes**, then edits the pinned message post-commit and best-effort. A 10:00 MSK cron job — the first recurring job in the project — collapses the day's unpublished changes to their net effect per date (five edits of one date become one line; a move and its undo drop out entirely), marks them published, writes the notifications in one transaction, and posts the digest. The same job unpins seasons whose last date has passed.

A game can be **linked** to a date: linking takes the date for the game's author if it was free and moves it onto the game's start, and a re-planned game drags its date along. Nothing is linked automatically — the engine offers, through a read-only `FindSlotsNearGameStartInteractor` both edges can call.

## Permissions

Promotion is the whole gate. Any author may compose, publish, edit and take — a date somebody else took included. Owning a date records who *means* to make the game there, not who may touch the row, so there is no `NotSlotOwner`, and taking a taken date is a re-assignment rather than a 409. `season_changes.actor_id` is what makes that recoverable, and there is an integration test asserting the trail names both people.

Two rules sit on top: naming a **team** as the author still needs its captain (`SlotAuthorInvalid`), and the named slot orgs may be any player, promoted or not. Putting *another* player or *another* team on a date is an admin-panel action — phase 4 — and is deliberately absent here: the ordinary schedule looks the same to an admin as to everyone else.

## Changes

- **Four new tables** — `seasons`, `season_slots`, `season_slot_orgs`, `season_changes` — plus their migration. No changes to existing tables.
- **`shvatka/core/season/`** — `dto`, `rules` (pure: default dates, the permission checks, the ±3-day search, digest collapsing), `adapters`, `interactors`. Per-table daos behind one composed `SeasonScheduleDao`, per the one-dao-per-interactor rule.
- **`SeasonAnnouncer`** (`core/views/season.py`) — a sibling of `GameLogWriter` that can hand a `message_id` back, which `GameLogWriter` cannot. `BotSeasonAnnouncer` sends, pins, edits in place and unpins; the web side is a no-op; `ComplexSeasonAnnouncer` fans out.
- **`/seasons` routes** — public to read, author-gated to write.
- **Glossary** rows in `context.md`, here and in bomzheg/shvatka-ui#203.

## Deviations from the SHEP

Recorded in the document's `Deviations` section: a linked date cannot be released (unlink first); `FindSlotsNearGameStartInteractor` takes a `window` defaulting to the ±3 days the offer uses; a date outside the season's year raises a bare `SeasonError` rather than a class nobody catches; `ComplexSeasonAnnouncer.publish` returns the bot's announcement rather than gathering one per edge.

On status codes, the SHEP asks for permission failures to reach 403. The api answers most `PermissionsError`s **422** today and `tests/integration/api_full/test_team.py` asserts that on purpose, with `# PermissionsError -> SHError` written beside it — so `CantBeAuthor` is 422 here too, and only `SlotAuthorInvalid` is 403. Making promotion failures 403 across the api is a worthwhile change, with those expectations to update alongside it, and not this one's to make.

## Testing

- **`tests/unit/season/`** — `default_slot_dates` across years including the 9-May-is-Saturday case; the permission rules; the ±3-day search including ties; digest collapsing; the loader options; and every interactor against an in-memory `SeasonScheduleDao`, which is what catches an interactor reading a change payload after the write that changes it.
- **`tests/integration/api_full/test_season.py`** — publish; publish the same year twice → 409; a date outside the year refused; reading without authentication; take; a player without promotion → 422; another author moves a taken date, and another author re-assigns it, with the trail naming them both; link a game → the date syncs; re-plan the game → the date follows; the suggest endpoint; the digest marks changes published and notifies; a finished season is unpinned once.
- **`tests/unit/test_api_router.py`** builds the whole route table, which is what catches a dataclass field named after its own type.

`ruff format`, `ruff check` and `mypy` (on a cold cache) are clean, and the unit suite — 716 tests, `test_di` included, which builds every container — passes locally. This sandbox has no Docker, so the testcontainer suite runs in CI.

https://claude.ai/code/session_01VFYy6TmPfofatvKwp4jNfP

---
_Generated by [Claude Code](https://claude.ai/code/session_01VFYy6TmPfofatvKwp4jNfP)_


shvatka/api/seasons/requests.py
```diff
@@ -0,0 +1,56 @@
...
+# a field named `date` shadows the type in the class namespace, and pydantic
+# evaluates the annotations there — so the type travels under its own name
+from datetime import date as date_
+
+from shvatka.core.season import dto
+
+
+@dataclass
+class SlotDraft:
+    date: date_
```
bomzheg: just make it slot_date. how do you think? or just date_ or at. i dont like to change imports when we have name conflict 

and other date names pls

bomzheg: Agreed, and `slot_date` of the three: `date_` would put a trailing underscore in the JSON that every client then copies, and `at` reads like a datetime next to `start_at` and `taken_at` while this is date-only.

Done in a34f9ad, and I took "other date names" to mean everywhere a slot's own day is the field — so both aliased imports are gone and the name is `slot_date` in all five places: the `season_slots` column, the orm attribute, `season.dto.Slot`/`SlotDraft`, the api request and the api response.

The wire contract moves with it (`{slot_date, note?}` for publish, add and edit), which costs nothing today since no client exists yet — phases 2 and 3 will be written against the new name. The SHEP is updated to match, including the trap entry, which used to say "import the type under another name" and now says rename the field.

One thing I left alone: the `"date"` key inside `season_changes.payload`. It is a dict key in JSONB, it shadows nothing, and it is the audit trail's stored format. Say the word if you want that renamed too.

`ruff`, `mypy` on a cold cache and the 716 unit tests are green.

---

shvatka/core/games/editor_interactors.py
```diff
@@ -164,13 +165,16 @@ class PlanGameStartInteractor:
     dao: GameStartPlanner
     scheduler: Scheduler
     game_log: GameLogWriter
+    slot_sync: SyncLinkedSlotInteractor
```
bomzheg: interactors cant use another interactor. your options:
- create core function (or even core class with __call__ for autowire with dishka. but do not call it interactor - it may be domain service or smth)
- send event to system bus

shvatka/core/interfaces/dal/season.py
```diff
@@ -0,0 +1,116 @@
...
+class SlotReader(Protocol):
+    async def get_slot(self, slot_id: int) -> dto.Slot:
+        raise NotImplementedError
+
+    async def get_slot_by_game(self, game_id: int) -> dto.Slot | None:
+        raise NotImplementedError
+
+    async def lock_slot(self, slot_id: int) -> dto.Slot:
+        """Re-read the row `FOR UPDATE`, so two takers cannot both win."""
+        raise NotImplementedError
```
bomzheg, line 57: i hope for this case it throw an error if try to lock and it is already locked. edit after another edit is a bad idea (it will still overwritten)

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,588 @@
...
+        if season is None:
+            raise exceptions.SeasonNotFound(text=f"no season for {year}")
+        return season
+
+    @staticmethod
+    def _get_slot(season: season_dto.Season, slot_id: int) -> season_dto.Slot:
+        for slot in season.slots:
+            if slot.id == slot_id:
+                return slot
+        raise exceptions.SlotNotFound(text=f"season {season.year} has no slot {slot_id}")
```
bomzheg: it should be method of season

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,588 @@
...
+    """What every season use case shares: the dao and the channel message."""
+
+    dao: SeasonScheduleDao
+    announcer: SeasonAnnouncer
+
+    async def _get_season(self, year: int) -> season_dto.Season:
+        season = await self.dao.get_season(year)
+        if season is None:
+            raise exceptions.SeasonNotFound(text=f"no season for {year}")
+        return season
```
bomzheg: just add to adapter get_required_season

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,588 @@
...
+
+class GetDefaultSlotDatesInteractor:
+    """The nine dates a season is composed from. Reads nothing, writes nothing."""
+
+    async def __call__(self, year: int) -> list[date]:
+        return default_slot_dates(year)
+
+
+@dataclass
+class SeasonInteractor:
```
bomzheg: i dont likee inheritance here. i guess we can just create some class helper to this and inject it

shvatka/infrastructure/db/dao/rdb/season.py
```diff
@@ -0,0 +1,249 @@
...
+            update(Season).where(Season.id == season_id).values(unpinned_at=self.clock(tz_utc))
+        )
+
+    async def touch_season(self, season_id: int) -> None:
+        await self.session.execute(
+            update(Season).where(Season.id == season_id).values(updated_at=self.clock(tz_utc))
+        )
+
+
+class SeasonSlotDao(BaseDAO[SeasonSlot]):
```
bomzheg: split daos by itself modules

shvatka/infrastructure/db/models/season.py
```diff
@@ -0,0 +1,219 @@
...
+            log_chat_id=self.log_chat_id,
+            log_message_id=self.log_message_id,
+            unpinned_at=self.unpinned_at,
+        )
+
+    def to_dto_with_slots(self) -> dto.Season:
+        return self.to_dto(slots=[slot.to_dto() for slot in self.slots])
+
+
+class SeasonSlot(Base):
```
bomzheg: split models by itself files

shvatka/core/games/editor_interactors.py
bomzheg: Took the first option. `SyncLinkedSlotInteractor` is now `LinkedSlotSync` in the new `shvatka/core/season/services.py` — a plain class with `__call__`, built by dishka like anything else, and not called an interactor.

The bus was the other candidate and I passed on it: the sync has to happen *before* `PlanGameStartInteractor` returns, because the date it moves is part of what the caller is about to be told, and an event would make that ordering implicit. A service keeps "the schedule follows the game" a rule the game's use case applies on purpose.

Done in ea6614f.

---

shvatka/core/interfaces/dal/season.py
bomzheg: You're right, and it did not — it waited. Fixed in ea6614f: the lock is `FOR UPDATE NOWAIT` now, and postgres refusing it (sqlstate 55P03) becomes `SlotIsBusy`, which the error handler maps to 409.

Exactly your reason, written into the SHEP: whoever queues behind an open edit decides against the row as it was *before* that edit and then writes their own version over it. Failing is the honest answer — the caller re-reads and decides again against what is actually there. Re-assigning a date on purpose stays allowed; doing it by accident does not.

Any other sqlstate travels on untouched, so a serialization failure still looks like one. `tests/unit/season/test_slot_locking.py` covers both branches.

---

shvatka/core/season/interactors.py
bomzheg: Done in ea6614f — it is `Season.get_slot(slot_id)` in `core/season/dto.py` now, raising `SlotNotFound` with the year in the message. It needed nothing but the season, so it had no business being a static method on an interactor.

---

shvatka/core/season/interactors.py
bomzheg: Done in ea6614f. `get_required_season(year)` is on `SeasonReader` beside `get_season`, implemented in `SeasonDao` and passed through `SeasonScheduleDaoImpl`, so every use case calls `self.dao.get_required_season(year)` and nobody unwraps the `None` any more.

---

shvatka/core/season/interactors.py
bomzheg: Agreed — the base class is gone in ea6614f. What was left of it after the two comments above became `ScheduleChangeLog` in `core/season/services.py`: `record()` writes the `season_changes` row inside the caller's transaction, `announce_update()` refreshes the pinned message after the commit. Injected, not inherited.

It made the real problem visible too. A base class hands every subclass the union of what any of them needs, so `GetSeasonInteractor` was carrying an announcer it never called. Now each use case's fields say exactly what it touches: a dao alone for the reads, `dao + changes` for the nine that edit, `dao + announcer` for publish and the digest job.

---

shvatka/infrastructure/db/dao/rdb/season.py
bomzheg: Done in ea6614f: `season.py` (`SeasonDao`), `season_slot.py`, `season_slot_org.py`, `season_change.py` — one class each, like every other dao in `rdb/`.

The shared loader options moved to `season_slot.py` as `SLOT_OPTIONS` (they describe what a slot walks into) and `season.py` imports them for its own `selectinload(Season.slots)`.

---

shvatka/infrastructure/db/models/season.py
bomzheg: Done in ea6614f: `season.py`, `season_slot.py`, `season_slot_org.py`, `season_change.py`, one model each. The cross-references are `TYPE_CHECKING` imports and SQLAlchemy's string-named relationships, so there is no import cycle; `configure_mappers()` is clean and `models/__init__.py` exports the same four names as before.

---

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,543 @@
...
+        await self.dao.commit()
+
+        published = await self.dao.get_required_season(year)
+        await self._announce_published(published)
+        await self._notify(
+            published,
+            actor=author,
+            payload={"year": year, "published": True, "slots": len(published.slots)},
+        )
+        return await self.dao.get_required_season(year)
```
bomzheg: it's the same as `published` above

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,543 @@
...
+        slot_id: int,
+        *,
+        author_kind: season_dto.SlotAuthorKind,
+        team_id: int | None,
+        org_player_ids: Collection[int],
+        identity: IdentityProvider,
+    ) -> season_dto.Slot:
+        actor = await identity.get_required_player()
+        season = await self.dao.get_required_season(year)
+        season.get_slot(slot_id)
```
bomzheg: do we need this? just get without variable

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,543 @@
...
+        orgs = [await self.dao.get_player_by_id(id_) for id_ in dict.fromkeys(org_player_ids)]
+        await self.dao.set_slot_orgs(slot_id, [org.id for org in orgs])
+        await self.changes.record(
+            season,
+            season_dto.ChangeType.slot_orgs_changed,
+            slot_id=slot_id,
+            actor=actor,
+            payload={
+                "date": slot.slot_date.isoformat(),
+                "orgs": [org.name_mention for org in orgs],
```
bomzheg, line 351: add here old orgs list pls

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,543 @@
...
+class LinkGameToSlotInteractor:
+    dao: SeasonScheduleDao
+    changes: ScheduleChangeLog
+
+    async def __call__(
+        self, year: int, slot_id: int, game_id: int, identity: IdentityProvider
+    ) -> season_dto.Slot:
+        actor = await identity.get_required_player()
+        season = await self.dao.get_required_season(year)
+        season.get_slot(slot_id)
```
bomzheg: lets add also method to season: ensure_have_slot(self, slot_id: int) -> None

shvatka/core/season/interactors.py
```diff
@@ -0,0 +1,543 @@
...
+        changes = await self.dao.get_unpublished_changes(season_id)
+        digests = collapse_changes(changes)
+        # every change is accounted for, including those that collapsed to nothing
+        await self.dao.mark_changes_published([change.id for change in changes])
+        if digests:
+            await self.dao.create_for_recipients(
+                recipient_ids=await self.dao.get_recipient_ids(_audience_since(now)),
+                type_=NotificationType.season_schedule_changed,
+                severity=NotificationSeverity.low,
+                payload={"year": season.year, "changes": len(digests)},
```
bomzheg: how web users should understand what changes was? we should show digests itself

shvatka/infrastructure/db/dao/complex/season.py
```diff
@@ -0,0 +1,196 @@
...
+            TeamPlayer.date_left.is_(None) | (TeamPlayer.date_left >= since)
+        )
+        organized = (
+            select(Organizer.player_id)
+            .join(Game, Game.id == Organizer.game_id)
+            .where(Organizer.deleted.is_(False), Game.start_at >= since)
+        )
+        authored = select(Game.author_id).where(Game.start_at >= since)
+        result = await self.dao.session.scalars(union(in_a_team, organized, authored))
+        return set(result.all())
```
bomzheg: please do not make query here. complex is always just proxy

shvatka/core/season/interactors.py
bomzheg: Done in 2cdcb31 — the payload carries `orgs_before` next to `orgs`, read before the write that replaces them (the same ordering trap the SHEP lists).

I carried it through to the digest too, otherwise it would be data nobody reads: `SlotDigest.orgs_before` keeps the list from the **first** orgs edit of the window, not the last, since a collapsed digest is the net change — two edits in one day should read «было @harry, стало @ron, @hermione», not «было @ron».

---

shvatka/core/season/interactors.py
bomzheg: Added in 2cdcb31: `Season.ensure_have_slot(slot_id) -> None`, used by the two callers that only need the id checked before they lock the row or read it back — `TakeSlotInteractor` and `LinkGameToSlotInteractor`. Both were discarding a whole `get_slot`, which reads like a mistake.

---

shvatka/core/season/interactors.py
bomzheg: You're right. I re-read it thinking the second query picked up what `_announce_published` wrote — it does, but `set_announcement` only sets `log_chat_id` and `log_message_id`, and `responses.Season` exposes neither (nor does it touch `updated_at`). So the two reads are identical to every caller, and it's just a wasted query.

Returns `published` now, in 2cdcb31.

---

shvatka/infrastructure/db/dao/complex/season.py
bomzheg: Moved in 2cdcb31. It is `PlayerDao.get_active_player_ids(since)` now, and `SeasonScheduleDaoImpl.get_recipient_ids` is a one-line forward like everything else there — the class has no SQLAlchemy import left at all.

`PlayerDao` because the query answers about players even though it reads `team_players`, `organizers` and `games` to find them, which is the same shape as `get_played_games_counts` next to it. I'd put my reasoning for keeping it in `complex/` on the record as wrong: I took "rooted at no single table" to mean it had nowhere else to go, when the right question was what it returns.

---

shvatka/core/season/interactors.py
bomzheg: Agreed — a count is a dead end. `season_changes` is the engine's audit trail, not an endpoint, so the feed item is the only place the web hears about any of this, and «3 изменения» leaves the reader with nowhere to go.

2cdcb31 sends the digests themselves: `SlotDigest.to_payload()` per date, dates as iso strings, and only the parts that actually changed, so a moved-and-taken date arrives as

```json
{"date": "2027-06-27", "moved_from": "2027-06-26", "moved_to": "2027-06-27", "owner": "@harry"}
```

and `notification-render.ts` renders «перенесена с 26.06 на 27.06; занял @harry» with no second fetch. The count is still there as `changes.length`, so the summary line the SHEP describes still works.

The SHEP's §Notification audience is updated to match, and phase 3's renderer will be written against this shape.

---

shvatka/infrastructure/db/models/season.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class Season(Base):
+    __tablename__ = "seasons"
+    __mapper_args__ = {"eager_defaults": True}
+
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    year: Mapped[int] = mapped_column(nullable=False, unique=True)
+    published_by_id: Mapped[int] = mapped_column(
+        ForeignKey("players.id", ondelete="CASCADE"), nullable=False
```
bomzheg: don't use cascades

shvatka/infrastructure/db/models/season_change.py
```diff
@@ -0,0 +1,59 @@
...
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    season_id: Mapped[int] = mapped_column(
+        ForeignKey("seasons.id", ondelete="CASCADE"), nullable=False
+    )
+    slot_id: Mapped[int | None] = mapped_column(
+        ForeignKey("season_slots.id", ondelete="SET NULL"), nullable=True
+    )
+    type: Mapped[str] = mapped_column(Text, nullable=False)
+    actor_id: Mapped[int | None] = mapped_column(
+        ForeignKey("players.id", ondelete="SET NULL"), nullable=True
```
bomzheg: remove ondelete here

shvatka/infrastructure/db/models/season_change.py
```diff
@@ -0,0 +1,59 @@
...
+
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    season_id: Mapped[int] = mapped_column(
+        ForeignKey("seasons.id", ondelete="CASCADE"), nullable=False
+    )
+    slot_id: Mapped[int | None] = mapped_column(
+        ForeignKey("season_slots.id", ondelete="SET NULL"), nullable=True
+    )
+    type: Mapped[str] = mapped_column(Text, nullable=False)
+    actor_id: Mapped[int | None] = mapped_column(
```
bomzheg: why nullable?

shvatka/infrastructure/db/models/season_slot.py
```diff
@@ -0,0 +1,102 @@
...
+    __mapper_args__ = {"eager_defaults": True}
+
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    season_id: Mapped[int] = mapped_column(
+        ForeignKey("seasons.id", ondelete="CASCADE"), nullable=False
+    )
+    slot_date: Mapped[date] = mapped_column(Date, nullable=False)
+    note: Mapped[str | None] = mapped_column(Text, nullable=True)
+    owner_id: Mapped[int | None] = mapped_column(
+        ForeignKey("players.id", ondelete="SET NULL"), nullable=True
```
bomzheg: remove ondelete here

shvatka/infrastructure/db/models/season_slot.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    author_kind: Mapped[dto.SlotAuthorKind | None] = mapped_column(
+        Enum(dto.SlotAuthorKind, name="slot_author_kind"), nullable=True
+    )
+    team_id: Mapped[int | None] = mapped_column(
+        ForeignKey("teams.id", ondelete="SET NULL"), nullable=True
```
bomzheg: remove ondelete

shvatka/infrastructure/db/models/season_slot.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    game_id: Mapped[int | None] = mapped_column(
+        ForeignKey("games.id", ondelete="SET NULL"), nullable=True
```
bomzheg: ok, remove all ondelte that you were written.

shvatka/infrastructure/db/models/season_slot.py
```diff
@@ -0,0 +1,102 @@
...
+        "Season", back_populates="slots", foreign_keys=[season_id]
+    )
+    owner: Mapped[models.Player | None] = relationship("Player", foreign_keys=[owner_id])
+    team: Mapped[models.Team | None] = relationship("Team", foreign_keys=[team_id])
+    game: Mapped[models.Game | None] = relationship("Game", foreign_keys=[game_id])
+    orgs: Mapped[list[SeasonSlotOrg]] = relationship(
+        "SeasonSlotOrg",
+        back_populates="slot",
+        foreign_keys="SeasonSlotOrg.slot_id",
+        cascade="all, delete-orphan",
```
bomzheg: no cascades!

shvatka/tgbot/views/season.py
```diff
@@ -0,0 +1,143 @@
...
+                message_id=season.log_message_id,
+                text=render_season(season),
+            )
+        except TelegramBadRequest as e:
+            if "message is not modified" not in str(e):
+                raise
+            logger.debug("schedule of %s is already up to date", season.year)
+
+    async def announce_digest(self, season: dto.Season, digests: Sequence[SlotDigest]) -> None:
+        await self.bot.send_message(
```
bomzheg, line 53: reply_to_message - season message itself

shvatka/views.py
```diff
@@ -77,6 +82,35 @@ async def log(self, log_event: GameLogEvent) -> None:
...
+
+    bot: BotSeasonAnnouncer
+    web: WebSeasonAnnouncer
+
+    async def publish(self, season: season_dto.Season) -> Announcement | None:
+        try:
+            await self.web.publish(season)
+        except Exception as e:
+            logger.exception("web season publish error", exc_info=e)
+        return await self.bot.publish(season)
```
bomzheg: why bot without try except?

shvatka/infrastructure/db/models/season_slot.py
bomzheg: All of them are gone in 48d6872 — every `ondelete=` in the four season models and in the migration, plus the ORM `cascade="all, delete-orphan"` on `SeasonSlot.orgs`. Nothing in the season schema cascades now, which matches the rest of the project.

Two places had been leaning on them, and both are explicit now:

**Removing a date.** `RemoveSlotInteractor` clears the orgs, detaches the change rows that point at the date (`SeasonChangeDao.detach_slot`), records `slot_removed` with no slot to point at, then deletes. The in-memory dao the interactor tests use asserts that order, so forgetting a step fails without a database, and `test_removing_a_taken_date_takes_its_orgs_and_keeps_its_trail` proves it against a real one.

**Merging two players** — the one place a player is really deleted. Without `ON DELETE SET NULL` the FK would refuse while a date still names them, so `replace_player_season` moves their dates, org seats and change rows to the survivor, beside the merge's other transfers. That's a better answer than what the cascade did anyway: nulling `owner_id` quietly freed the date with nothing in the trail to say why. A merged player who was already an org on the same date isn't seated twice — the duplicate `(slot_id, player_id)` row is dropped rather than collided with.

I'd flag the merge part as the one bit of this that reaches outside the season area (`core/players/player.py`, the `PlayerMerger` protocol, `complex/player.py`). It's four lines plus three per-table writes, and leaving it out would mean a merge crashing on a foreign key in production, so I did it rather than leave the defect — but say the word if you'd rather it were its own PR.

---

shvatka/infrastructure/db/models/season_change.py
bomzheg: Because a change row has to outlive the date it describes — it's the audit trail, and the digest renders from `payload`, which denormalizes the date, owner, team, game and org names precisely so a removed date still reads correctly weeks later.

That was also the reason for the `ON DELETE SET NULL` you've just had me remove. It's explicit now: `RemoveSlotInteractor` calls `detach_slot` before deleting, and records `slot_removed` with `slot_id=None` from the start. So the column stays nullable, but nothing nulls it behind your back.

The alternative would be a non-nullable `slot_id` and deleting the trail with the date, which throws away the one thing that made the payload worth denormalizing. I've written the reason into the SHEP's data model next to the column so it doesn't get asked again.

---

shvatka/tgbot/views/season.py
bomzheg: Done in 48d6872 — `announce_digest` passes `reply_to_message_id=season.log_message_id`, so the day's lines hang off the pinned schedule and a reader of «перенесена с 15.05 на 17.05» can reach the plan it changed.

It's `None` when nothing was announced (no pin right, or telegram refused the post), which telegram treats as an ordinary message rather than an error — the same fallback `announce_digest` already uses for the chat id.

---

shvatka/views.py
bomzheg: No good reason — it was an oversight. I wrapped the web call and then let the bot call through because it carries the return value, without noticing that made the two edges behave differently in exactly the case the try/except exists for.

Wrapped in 48d6872: a failed publish logs and returns `None`, which the interactor already handles — no announcement back means no message id stored, and `update` no-ops until one exists. So a channel that refuses the post can't take a published season down with it, which is what the SHEP says should happen and what `ComplexGameLogWriter` does on both sides.

---

shvatka/infrastructure/db/models/season.py
bomzheg: Gone in 48d6872 — see the [consolidated reply](https://github.com/bomzheg/Shvatka/pull/417#discussion_r4044169705) for what replaced the cascades.

---

shvatka/infrastructure/db/models/season_change.py
bomzheg: Gone in 48d6872 — see the [consolidated reply](https://github.com/bomzheg/Shvatka/pull/417#discussion_r4044169705).

---

shvatka/infrastructure/db/models/season_slot.py
bomzheg: Gone in 48d6872 — see the [consolidated reply](https://github.com/bomzheg/Shvatka/pull/417#discussion_r4044169705).

---

shvatka/infrastructure/db/models/season_slot.py
bomzheg: Gone in 48d6872 — see the [consolidated reply](https://github.com/bomzheg/Shvatka/pull/417#discussion_r4044169705).

---

shvatka/infrastructure/db/models/season_slot.py
bomzheg: Gone in 48d6872 — this was the ORM `cascade="all, delete-orphan"`, and it was misleading on top of being unwanted: `remove_slot` issues a bulk `delete()`, which never triggered it. The orgs are cleared explicitly now. See the [consolidated reply](https://github.com/bomzheg/Shvatka/pull/417#discussion_r4044169705).

---

